PR #1, opened 2023-03-24
Add README.md and datasets/ directory.

pierrot0:
I think than soon we want a directory for small artificial examples (as opposed to full datasets) which will be useful for both documentation and tests. I suppose we should also have a directory in which we can define the schema ml types, and then one or several directories for the tools / libraries we create. Here maybe something like "python/" (in case more tools in other language are needed later) and in there "lib/" and "bin/" maybe.


pierrot0:
recheck

nathanw-mlc:
recheck

PR #2, opened 2023-04-04
Add titanic dataset config

pierrot0:
This config does not include semantic information, such as for example how to interpret the gender, the embarkation port, etc. This will come later.

Considerations which might be worth a discussion:
 - The quality of the data really could be improved. I have added some information in the description but maybe we can come up with some more precise annotations.
 - Aliases: when naming the fields in a recordSet, would it make sense to allow for aliases? In titanic original dataset, some field names aren't very explicit (eg: "parch"), so I renamed them (eg: "num_parents_children"). That's discutable, but an alias would allow to keep both.

### DIFF
--- a/datasets/titanic/metadata.json
+++ b/datasets/titanic/metadata.json
@@ -1,0 +1,141 @@
+{
+    "@context": {
+      "@vocab": "https://schema.org/",
+      "ml": "http://mlcommons.org/schema/"
+    },
+  
+    "@type": "Dataset",
+    "@language": "en",
+    "name": "Titanic",
+    "url": "https://www.openml.org/d/40945",
+    "description": "The original Titanic dataset, describing the status of individual passengers on the Titanic.\n\n The titanic data does not contain information from the crew, but it does contain actual ages of half of the passengers. \n\n For more information about how this dataset was constructed: \nhttps://web.archive.org/web/20200802155940/http://biostat.mc.vanderbilt.edu/wiki/pub/Main/DataSets/titanic3info.txt\n\nOther useful information (useful for prices description for example):\nhttp://campus.lakeforest.edu/frank/FILES/MLFfiles/Bio150/Titanic/TitanicMETA.pdf\n\n Also see the following article describing shortcomings of the dataset data:\nhttps://emma-stiefel.medium.com/plugging-holes-in-kaggles-titanic-dataset-an-introduction-to-combining-datasets-with-fuzzywuzzy-60a686699da7\n",
+    "license": "Public",
+    "citation": "The principal source for data about Titanic passengers is the Encyclopedia Titanica (http://www.encyclopedia-titanica.org/). The datasets used here were begun by a variety of researchers. One of the original sources is Eaton & Haas (1994) Titanic: Triumph and Tragedy, Patrick Stephens Ltd, which includes a passenger list created by many researchers and edited by Michael A. Findlay.\n\nThomas Cason of UVa has greatly updated and improved the titanic data frame using the Encyclopedia Titanica and created the dataset here. Some duplicate passengers have been dropped, many errors corrected, many missing ages filled in, and new variables created.\n",
+  
+    "distribution": [
+      {
+        "@type": "FileObject",
+        "name": "passengers",
+        "contentUrl": "https://www.openml.org/data/get_csv/16826755/phpMYEkMl",
+        "sha256": "c617db2c7470716250f6f001be51304c76bcc8815527ab8bae734bdca0735737",
+        "encodingFormat": "text/csv"
+      }
+    ],
+  
+    "recordSet": [
+      {
+        "@type": "ml:RecordSet",
+        "name": "passengers_list",
+        "description": "The list of passengers. Does not include crew members.\n\nThere seem to be at least two passengers with the same name, so a composite key is used to uniquely identify passengers",
+        "source": "#{passengers}",
+        "key": [
+          "#{name}",
+          "#{age}",
+          "#{gender}",
+          "#{embarked}",
+          "#{home_destination}"
+        ],
+        "field": [
+          {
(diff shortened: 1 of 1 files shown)


datasets/titanic/metadata.json
```diff
@@ -0,0 +1,141 @@
...
+          "#{gender}",
+          "#{embarked}",
+          "#{home_destination}"
+        ],
+        "field": [
+          {
+            "name": "name",
+            "description": "Name of the passenger",
+            "@type": "ml:Field",
+            "dataType": "Text",
```
josvandervelde, line 43: In the DCF spec document, it is stated that `dataType` "builds on the Datatype class in [CSVW](https://csvw.org/)."
I'm new to CSVW, but I found this image with all possible CSVW types: https://www.w3.org/TR/tabular-data-primer/datatypes.png (found on https://www.w3.org/TR/tabular-data-primer/). 

If this source is correct, it should be `string` not `Text`, `integer` not `Integer`, `float` not `Float`.

datasets/titanic/metadata.json
```diff
@@ -0,0 +1,141 @@
...
+      }
+    ],
+  
+    "recordSet": [
+      {
+        "@type": "ml:RecordSet",
+        "name": "passengers_list",
+        "description": "The list of passengers. Does not include crew members.\n\nThere seem to be at least two passengers with the same name, so a composite key is used to uniquely identify passengers",
+        "source": "#{passengers}",
+        "key": [
```
benjelloun, line 31: Do we need all these fields to be part of the key, or would a subset be sufficient? E.g., (name, age)

benjelloun:
Re aliases: Schema.org supports them via the https://schema.org/alternateName property.

benjelloun (approved):

datasets/titanic/metadata.json
```diff
@@ -0,0 +1,141 @@
...
+    "license": "Public",
+    "citation": "The principal source for data about Titanic passengers is the Encyclopedia Titanica (http://www.encyclopedia-titanica.org/). The datasets used here were begun by a variety of researchers. One of the original sources is Eaton & Haas (1994) Titanic: Triumph and Tragedy, Patrick Stephens Ltd, which includes a passenger list created by many researchers and edited by Michael A. Findlay.\n\nThomas Cason of UVa has greatly updated and improved the titanic data frame using the Encyclopedia Titanica and created the dataset here. Some duplicate passengers have been dropped, many errors corrected, many missing ages filled in, and new variables created.\n",
+  
+    "distribution": [
+      {
+        "@type": "FileObject",
+        "name": "passengers",
+        "contentUrl": "https://www.openml.org/data/get_csv/16826755/phpMYEkMl",
+        "sha256": "c617db2c7470716250f6f001be51304c76bcc8815527ab8bae734bdca0735737",
```
josvandervelde, line 20: `sha256` is not mentioned in the DCF spec, and also not part of https://schema.org/CreativeWork

josvandervelde (commented):
Looks good to me!

As a reference, I generated a `titanic` example in https://github.com/mlcommons/datasets_format/blob/feature/openml-croissant-converter/datasets/Titanic/openml_titanic.json.

Main difference is, I think, that I generated it automatically, so it does not have some improvements:
- field naming -> I used the exact names as in OpenML
- field descriptions -> Not present in OpenML, so I did not include them
But it does have some extra fields such as `dateCreated`.

I wanted to discuss the generated OpenML schemas in the meeting this afternoon

datasets/titanic/metadata.json
pierrot0: I think the idea would be to reuse the https://schema.org/sha256 property on FileObject rdfs:class, which doesn't exist yet AFAIK. @benjelloun do you know if the FileObject can already be seen somewhere?

datasets/titanic/metadata.json
pierrot0: Thanks, you are right, fixed.

datasets/titanic/metadata.json
benjelloun: I think it's fine to omit the key in that case.

pierrot0:
> Re aliases: Schema.org supports them via the https://schema.org/alternateName property.

That's cool!
Let's also talk during the meeting if that is something we want to support right now as part of the config of if this is something we want to delay the implementation.

pierrot0:
> > Re aliases: Schema.org supports them via the https://schema.org/alternateName property.
> 
> That's cool! Let's also talk during the meeting if that is something we want to support right now as part of the config of if this is something we want to delay the implementation.

Following VS conversation: let's not use alternateName for now. We can always refer to the original column name if needed too.

datasets/titanic/metadata.json
```diff
@@ -0,0 +1,141 @@
...
+    "license": "Public",
+    "citation": "The principal source for data about Titanic passengers is the Encyclopedia Titanica (http://www.encyclopedia-titanica.org/). The datasets used here were begun by a variety of researchers. One of the original sources is Eaton & Haas (1994) Titanic: Triumph and Tragedy, Patrick Stephens Ltd, which includes a passenger list created by many researchers and edited by Michael A. Findlay.\n\nThomas Cason of UVa has greatly updated and improved the titanic data frame using the Encyclopedia Titanica and created the dataset here. Some duplicate passengers have been dropped, many errors corrected, many missing ages filled in, and new variables created.\n",
+  
+    "distribution": [
+      {
+        "@type": "FileObject",
+        "name": "passengers",
```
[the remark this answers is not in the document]
pierrot0: Works for me, I think we should have guidelines here on how to name recordSets, files, etc.

datasets/titanic/metadata.json
pierrot0: Removed.

PR #3, opened 2023-04-06
Feature/openml croissant converter

josvandervelde:
Birds and Titanic dataset metadata in DCF format, including the converter-code.
I changed the points we discussed last meeting (e.g. datatype from csvw to Schema.org).

Open TODOs:
- add `possible_values` for categorical variables
- add `checksum` to `FileObject`
- add `sameAs` to `FileObject`

I think that the resulting json files are the most important part to review.

Point to discuss:
- `distribution.name` should be human readable I think: "[dataset name] ([format])", e.g. "titanic (parquet)". But it's awkward to reference this:  `"#{Titanic (parquet)/name}"`. `"#{titanic_parquet/name}"` would be better, but then we have to change the `distribution.name`, replacing all special characters. Do we want that?


josvandervelde:
Hi @benjelloun , @pierrot0 and @PGijsbers . I assigned you to this PR, but no problem if you don't have time / I should assign others. Just let me know!
Any feedback is appreciated, most important would probably be resulting jsons.
Thanks!

PGijsbers:
I will be on two weeks leave and won't have time to process this before that. If anyone else wants to pick this up, that would be great!

python/openmlconverter/src/converter.py
```diff
@@ -0,0 +1,151 @@
...
+    _remove_none_values(croissant)
+    for recordSet in croissant["recordSet"]:
+        _remove_none_values(recordSet)
+    return croissant
+
+
+def _get_field(json_dict, field: str, transform=None, required=False):
+    """
+    Convenience function. Get field and optionally perform the transformation. If field does not
+    exist, throw an error if the field is  required, otherwise return None.
```
benjelloun, line 67: Please document the function arguments and return value as well.

python/openmlconverter/src/test/resources/dcf/openml_2_expected.json
benjelloun, line 1: Maybe the test should be run on a constructed example that exercises all the features instead of a full actual dataset?

benjelloun (approved):
Looks good overall ! A few small comments.

(I'm not a Python expert, so someone else may want to review the code as well.)

datasets/birds/birds_by_openml_converter.json
```diff
@@ -0,0 +1,1715 @@
...
+  "creativeWorkStatus": "active",
+  "dateCreated": "2019-03-14T17:58:26",
+  "dateModified": "2019-03-14T17:59:14",
+  "description": "Multi-label dataset. The birds dataset consists of 327 audio\n  recordings of 12 different vocalizing bird species. Each sound can be\n  assigned to various bird species.",
+  "distribution": [
+    {
+      "@type": "FileObject",
+      "contentUrl": "http://openml1.win.tue.nl/dataset41464/dataset_41464.pq",
+      "encodingFormat": "application/vnd.apache.parquet",
+      "name": "birds (parquet)"
```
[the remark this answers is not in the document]
josvandervelde: Thanks @benjelloun for the review! 
I will fix these points coming Wednesday

datasets/titanic/titanic_by_openml_converter.json
```diff
@@ -0,0 +1,124 @@
...
+  "distribution": [
+    {
+      "@type": "FileObject",
+      "contentUrl": "http://openml1.win.tue.nl/dataset40945/dataset_40945.pq",
```
pierrot0, line 15: It would be nice to add a checksum (sha256).
This is critical for libraries such as TFDS who want to verify the integrity of files before processing them, both from a security and correctness perspective.

We should also set [contentSize](https://schema.org/contentSize) (I also haven't set it in my previous CL), as this would make it easier to show how much download to expect, download progress, etc.

datasets/titanic/titanic_by_openml_converter.json
```diff
@@ -0,0 +1,124 @@
...
+          "dataType": "Text",
+          "name": "home.dest",
+          "source": "#{Titanic (parquet)/home.dest}"
+        }
+      ],
+      "name": "Titanic",
+      "source": "#{Titanic (parquet)}"
+    }
+  ],
+  "url": "https://www.openml.org/api/v1/json/data/40945",
```
pierrot0, line 122: This is a JSON containing roughly the same data as specified in other fields, like description, right?
Should we have some guidelines on what this URL should prefer pointing at? Machine readable vs human friendly URL? The later with schema.org embedded machine readable info?

@benjelloun

python/openmlconverter/src/converter.py
pierrot0: At Google this is the docstring formatting we use: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings

I have no strong feelings about using one or another format, as long as we are consistent within each sub-project.

python/openmlconverter/src/converter.py
```diff
@@ -0,0 +1,151 @@
...
+        mimetype = "application/vnd.apache.parquet"  # Not an official mimetype yet
+        # see https://issues.apache.org/jira/browse/PARQUET-1889
+    else:
+        raise ValueError(f"Unrecognized file extension in url: {url}")
+    return {
+        "name": f"{name} ({type_})",
+        "@type": "FileObject",
+        "contentUrl": url,
+        "encodingFormat": mimetype
+        # TODO: add the md5 hash? But where? 'sha256' is in the Google Docs file, but not official?
```
pierrot0, line 103: https://schema.org/sha256 is pending implementation feedback but I suppose we can use it.
TFDS also uses MD5, and while re-computing the sha256 is possible, I don't know how much friction it would add to the various actors...

python/openmlconverter/src/converter.py
```diff
@@ -0,0 +1,151 @@
...
+        # TODO: add sameAs (other distribution)?
```
pierrot0, line 104: So that would be for proxies serving the same file?

python/openmlconverter/src/converter.py
```diff
@@ -0,0 +1,151 @@
...
+        return row_identifiers[0]
+    return None
+
+
+def _lenient_date_parser(value: str) -> datetime.date | datetime.datetime:
+    try:
+        dateutil.parser.parse(value)
+    except dateutil.parser.ParserError:
+        pass
+    if len(value) == 4 and (value.startswith("19") or value.startswith("20")):
```
pierrot0, line 138: I would remove `and (value.startswith("19") or value.startswith("20")`: if we are in this function, then we already know that's supposed to be a date, no?
Maybe you'll have a dataset which an older book published before 1900?

pierrot0:
Thank you, LGTM overall, just left a few comments.

pierrot0 (approved):

datasets/birds/birds_by_openml_converter.json
```diff
@@ -0,0 +1,1715 @@
...
+          "dataType": "Boolean",
+          "name": "Common.Nighthawk",
+          "source": "#{birds (parquet)/Common.Nighthawk}"
+        }
+      ],
+      "name": "birds",
+      "source": "#{birds (parquet)}"
+    }
+  ],
+  "url": "https://www.openml.org/api/v1/json/data/41464",
```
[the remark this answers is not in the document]
josvandervelde: Fixed: https://github.com/mlcommons/datasets_format/pull/3/commits/b8fc0cf5714526ef795511d4feb4853ac30acd98

python/openmlconverter/README.md
```diff
@@ -0,0 +1,26 @@
+# OpenML to MLCommons converter 
+
+This repository contains the code to convert from the OpenML dataset schema to the new MLCommons
```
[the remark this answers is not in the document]
josvandervelde: Fixed: https://github.com/mlcommons/datasets_format/pull/3/commits/8721320a54d39ab277c88b1a2144747ab541cb56

python/openmlconverter/src/converter.py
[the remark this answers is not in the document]
josvandervelde: Fixed: https://github.com/mlcommons/datasets_format/pull/3/commits/a83b64ba1f9fe2beec1bac970fd0354c4c10d94b

python/openmlconverter/src/converter.py
```diff
@@ -0,0 +1,151 @@
+import datetime
+from functools import partial
+
+import dateutil.parser
+import langcodes
+
+
+def bake(openml_dataset: dict, openml_features: dict) -> dict:
```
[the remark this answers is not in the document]
josvandervelde: Fixed: https://github.com/mlcommons/datasets_format/pull/3/commits/a83b64ba1f9fe2beec1bac970fd0354c4c10d94b

python/openmlconverter/src/converter.py
josvandervelde: Thanks! I used the Google docstring formatting.
For this file: https://github.com/mlcommons/datasets_format/pull/3/commits/a83b64ba1f9fe2beec1bac970fd0354c4c10d94b
For other files: https://github.com/mlcommons/datasets_format/pull/3/commits/40e89fe17fed52629b9d235dd14f7dfde2c59751

python/openmlconverter/src/converter.py
```diff
@@ -0,0 +1,151 @@
...
+            }
+        ],
+    }
+    _remove_none_values(croissant)
+    for recordSet in croissant["recordSet"]:
+        _remove_none_values(recordSet)
+    return croissant
+
+
+def _get_field(json_dict, field: str, transform=None, required=False):
```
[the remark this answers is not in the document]
josvandervelde: Fixed! https://github.com/mlcommons/datasets_format/pull/3/commits/1d1adcb1b268e70e9ecdfc12d445cc35c63e27d8

python/openmlconverter/src/converter.py
josvandervelde: In OpenML we currently have 2 files with the same content. An `arff` file and a `parquet` file. It would be nice to add a `sameAs` to both distributions.

For OpenML, this is not necessary, we're migrating towards parquet-only. But in general, it would be nice.

python/openmlconverter/src/converter.py
josvandervelde: I understand your point, but personally I like to throw an exception in such weird cases. I'd like to look at a couple such examples. I'd suggest keeping the Exception and removing it after seeing an instance such as 1899 which is actually correct (and not a value such as 1234). I'm a bit afraid of the content of this field. It should be only dates, but the content format is not very strict.

What do you think?

python/openmlconverter/src/test/resources/dcf/openml_2_expected.json
josvandervelde: Fixed! https://github.com/mlcommons/datasets_format/pull/3/commits/252e74b0a586227c88f2368a7cb6e91066bf9c8b

pierrot0:
Works for me!

python/openmlconverter/src/converter.py
pierrot0: Works for me.

python/openmlconverter/src/converter.py
pierrot0: Agreed, there are a few examples of datasets with multiple mirrors where to download the data from.

How about we create a new issue (https://github.com/mlcommons/datasets_format/issues/new) for each of such topics, and then we discuss them on the issue or during our meetings?

datasets/birds/birds_by_openml_converter.json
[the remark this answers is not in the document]
josvandervelde: Great, I fixed this in https://github.com/mlcommons/datasets_format/pull/3/commits/2ce4c0a365a3c4cb998949efbfc25a6dc1040a7c.
I also added a bit of text in the DCF spec.

python/openmlconverter/src/converter.py
josvandervelde: Good idea! I created https://github.com/mlcommons/datasets_format/issues/7. Let's discuss this issue, and the general usage of issues, in the next meeting.

datasets/titanic/titanic_by_openml_converter.json
josvandervelde: Sounds good! I created a separate issue (https://github.com/mlcommons/datasets_format/issues/8) because I need to do some checks with OpenML, and maybe improve the OpenML REST API a bit to return the correct information.

python/openmlconverter/src/converter.py
josvandervelde: I resolve this conversation for now, moved it to a separate issue: https://github.com/mlcommons/datasets_format/issues/8

datasets/titanic/titanic_by_openml_converter.json
josvandervelde: Good point! I changed the URL to a human friendly page that describes the dataset. This is also more in line with https://developers.google.com/search/docs/appearance/structured-data/dataset, where they describe the `url` as "Location of a page describing the dataset.".

Fixed in https://github.com/mlcommons/datasets_format/pull/3/commits/1a81c16e3a9bbe96471da88ccd91f6bc24e31dc9

PR #4, opened 2023-04-11
Add drive folder to README.md

pierrot0:


benjelloun:
Should we make the drive folder publicly accessible?

pierrot0:
I think we should make the drive folder publicly readable indeed.

PR #5, opened 2023-04-11
Add contentSize to titanic FileObject

pierrot0:
https://schema.org/contentSize is not very explicit on how the size unit should be specified.
https://github.com/schemaorg/schemaorg/issues/869 is the corresponding issue on schema.org.

### DIFF
--- a/datasets/titanic/metadata.json
+++ b/datasets/titanic/metadata.json
@@ -18,6 +18,7 @@
         "@type": "FileObject",
         "name": "passengers-table",
         "contentUrl": "https://www.openml.org/data/get_csv/16826755/phpMYEkMl",
+        "contentSize": "117743B",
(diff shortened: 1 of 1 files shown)


datasets/titanic/metadata.json
```diff
@@ -18,6 +18,7 @@
         "@type": "FileObject",
         "name": "passengers-table",
         "contentUrl": "https://www.openml.org/data/get_csv/16826755/phpMYEkMl",
+        "contentSize": "117743B",
```
benjelloun, line 21: Nit: Would it make sense to have a space between the number and the B?

benjelloun (approved):
Yeah, contentSize is underspecified unfortunately.

pierrot0:
> 

Starting discussion on the spec doc directly: https://docs.google.com/document/d/1EDUbwwSeSUpRZWczy6FNuCXTbKh9e5H4xXMnqgoCRj8/edit?disco=AAAAvCo8OuU

PR #6, opened 2023-04-12
Titanic dataset: add enum semantic definitions for gender and embarkation port, using inlined data

pierrot0:
This change defines two new `recordSet`s:
 1. mapping from gender (0, 1) to labels ("Male", "Female") and their semantic meaning as a wikidata url (eg: https://www.wikidata.org/wiki/Q6581097).
 2. mapping from embarkation port ("C", "Q", "S", "?") to labels and their semantic meaning as a wikidata url. This is a good example of wikidata benefits, as "Queenstown" doesn't refer to the New-Zealand city of that name, but to the Irish city which has since then been renamed Cobh.

The meta semantic types (or classes) - gender and city - are defined together with the dataType. I think we should see this feature as orthogonal to the semantic type: one could be defined without the other, and it would be acceptable I think to have a dataset where a field has both data types picture and url (eg: https://www.wikidata.org/wiki/Q506).

The two new recordsets have their data defined inline, using a simple json notation.

What I like about this change is that the semantic definition of the gender and embarkation_port fields is fairly independent of the data passengers data definition, so one can easily add it at a later stage of the dataset definition...

Other small changes: 
 - Import https://schema.org/ under "sc:" namespace, so it's more obvious when reading the config what comes from there.
 - Switch back from http://www.w3.org/ns/csvw to schema.org for Text and Integer dataTypes, as per https://github.com/mlcommons/datasets_format/commit/bf3813c7d210b0ffaa062382408b9a29c6a0d80e#comments.

### DIFF
--- a/datasets/titanic/metadata.json
+++ b/datasets/titanic/metadata.json
@@ -1,21 +1,21 @@
 {
     "@context": {
-      "@vocab": "https://schema.org/",
-      "csvw": "http://www.w3.org/ns/csvw",
-      "ml": "http://mlcommons.org/schema/"
+      "sc": "https://schema.org/",
+      "ml": "http://mlcommons.org/schema/",
+      "wd": "https://www.wikidata.org/wiki/Q"
     },
-  
-    "@type": "Dataset",
+
+    "@type": "sc:Dataset",
     "@language": "en",
     "name": "Titanic",
     "url": "https://www.openml.org/d/40945",
     "description": "The original Titanic dataset, describing the status of individual passengers on the Titanic.\n\n The titanic data does not contain information from the crew, but it does contain actual ages of half of the passengers. \n\n For more information about how this dataset was constructed: \nhttps://web.archive.org/web/20200802155940/http://biostat.mc.vanderbilt.edu/wiki/pub/Main/DataSets/titanic3info.txt\n\nOther useful information (useful for prices description for example):\nhttp://campus.lakeforest.edu/frank/FILES/MLFfiles/Bio150/Titanic/TitanicMETA.pdf\n\n Also see the following article describing shortcomings of the dataset data:\nhttps://emma-stiefel.medium.com/plugging-holes-in-kaggles-titanic-dataset-an-introduction-to-combining-datasets-with-fuzzywuzzy-60a686699da7\n",
     "license": "Public",
     "citation": "The principal source for data about Titanic passengers is the Encyclopedia Titanica (http://www.encyclopedia-titanica.org/). The datasets used here were begun by a variety of researchers. One of the original sources is Eaton & Haas (1994) Titanic: Triumph and Tragedy, Patrick Stephens Ltd, which includes a passenger list created by many researchers and edited by Michael A. Findlay.\n\nThomas Cason of UVa has greatly updated and improved the titanic data frame using the Encyclopedia Titanica and created the dataset here. Some duplicate passengers have been dropped, many errors corrected, many missing ages filled in, and new variables created.\n",
-  
+
     "distribution": [
       {
-        "@type": "FileObject",
+        "@type": "sc:FileObject",
         "name": "passengers-table",
         "contentUrl": "https://www.openml.org/data/get_csv/16826755/phpMYEkMl",
         "contentSize": "117743 B",
@@ -23,8 +23,48 @@
         "encodingFormat": "text/csv"
       }
     ],
-  
+
     "recordSet": [
+      {
+        "@type": "ml:RecordSet",
+        "name": "gender_enums",
+        "description": "Maps gender keys (0, 1) to labeled values.",
+        "key": "#{key}",
+        "field": [
+            { "name": "key", "@type": "ml:Field", "dataType": "sc:Integer" },
+            { "name": "label", "@type": "ml:Field", "dataType": "sc:String" },
+            {
+                "name": "url",
+                "@type": "ml:Field",
+                "dataType": ["sc:Url", "wd:48277"]
+            }
+        ],
+        "data": [
+            { "key": 0, "label": "Male", "url": "wd:6581097" },
+            { "key": 1, "label": "Female", "url": "wd:6581072" }
+        ]
+      },
+      {
+        "@type": "ml:RecordSet",
+        "name": "ports_enums",
+        "description": "Maps Embarkation port initial to labeled values.",
+        "key": "#{key}",
+        "field": [
+            { "name": "key", "@type": "ml:Field", "dataType": "sc:Text" },
+            { "name": "label", "@type": "ml:Field", "dataType": "sc:Text" },
+            {
+                "name": "url",
+                "@type": "ml:Field",
+                "dataType": ["sc:Url", "wd:515"]
+            }
+        ],
+        "data": [
+            { "key": "C", "label": "Cherbourg", "url": "wd:3667188" },
+            { "key": "Q", "label": "Queenstown", "url": "wd:733093" },
+            { "key": "S", "label": "Southampton", "url": "wd:79848" },
+            { "key": "?", "label": "Unknown", "url": "wd:24238356" }
+        ]
+      },
       {
         "@type": "ml:RecordSet",
         "name": "passengers",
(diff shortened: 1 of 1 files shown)


datasets/titanic/metadata.json
```diff
@@ -1,30 +1,70 @@
...
         "sha256": "c617db2c7470716250f6f001be51304c76bcc8815527ab8bae734bdca0735737",
         "encodingFormat": "text/csv"
       }
     ],
-  
+
     "recordSet": [
+      {
+        "@type": "ml:RecordSet",
+        "name": "gender_enums",
```
josvandervelde, line 30: I like it! Just some thoughts:
1) We should update the DCF spec

2) Should we have an alternative for a simpler enum? When automatically generating this a DCF, we normally only know the values, not any meaning. I think that it would be nice to do something like:
```{ "@type": "ml:RecordSet", "name": "gender_enums", "data": [0, 1]}```

File: datasets/titanic/metadata.json [hunk 2]
```diff
@@ -1,30 +1,70 @@
...
+        "field": [
+            { "name": "key", "@type": "ml:Field", "dataType": "sc:Integer" },
+            { "name": "label", "@type": "ml:Field", "dataType": "sc:String" },
+            {
+                "name": "url",
+                "@type": "ml:Field",
+                "dataType": ["sc:Url", "wd:48277"]
+            }
+        ],
+        "data": [
```
josvandervelde, line 42: `data` is a new field, right? I could not find a similar field in other schema.org types.
If it's not an existing field, would it make sense to be more descriptive here, such as `possible_value` instead of `data`? Or do we see other uses for this field?

datasets/titanic/metadata.json
```diff
@@ -1,30 +1,70 @@
...
+        "name": "gender_enums",
+        "description": "Maps gender keys (0, 1) to labeled values.",
+        "key": "#{key}",
+        "field": [
+            { "name": "key", "@type": "ml:Field", "dataType": "sc:Integer" },
+            { "name": "label", "@type": "ml:Field", "dataType": "sc:String" },
+            {
+                "name": "url",
+                "@type": "ml:Field",
+                "dataType": ["sc:Url", "wd:48277"]
```
josvandervelde, line 39: Alternatively we could have two fields: `semanticType` next to `dataType`, highlighting the different meaning of both values, at the cost of adding another field.
What do you think?

datasets/titanic/metadata.json
```diff
@@ -35,51 +75,52 @@
...
             "name": "embarked",
             "description": "Port of Embarkation (C: Cherbourg, Q: Queenstown, S: Southampton, ?: Unknown).",
-            "@type": "City/name",
-            "dataType": "csvw:string",
-            "source": "#{passengers-table/embarked}"
-  
+            "@type": "ml:Field",
+            "dataType": "sc:Text",
+            "source": "#{passengers-table/embarked}",
+            "references": "#{ports_enums}"
```
josvandervelde, line 123: See earlier comment about `semanticType`. We could change `references` into `semanticType` here.
`"semanticType": "#{ports_enum}"` might be more descriptive. I think.

josvandervelde (commented):
I like this approach! Added some thoughts to discuss :)

datasets/titanic/metadata.json
pierrot0: 1. Definitely, as soon as this pull request is merged, I'll add.

2. In TFDS,  for Titanic and many other datasets, we should be able to generate the DCF config with the labels:
```
      {
        "@type": "ml:RecordSet",
        "name": "gender_enums",
        "description": "List gender values.",
        "key": "#{key}",
        "field": [
            { "name": "key", "@type": "ml:Field", "dataType": "sc:Integer" },
            { "name": "label", "@type": "ml:Field", "dataType": "sc:String" }
        ],
        "data": [
            { "key": 0, "label": "Male" },
            { "key": 1, "label": "Female" }
        ]
      }
```

Generating an incomplete gender_enums Recordset would minimize the size of the next diff and make it easier to later add the wikidata IDs.

If there is not enough information to have the labels and this is just about enumerating the different possible enum values, then I would drop the additional Recordset all together. Something like:

```
          {
            "name": "gender",
            "description": "Gender of passenger (0: Male, 1: Female)",
            "@type": "ml:Field",
            "dataType": "sc:Integer",
            "possibleValues": [0, 1],
            "source": "#{passengers-table/sex}",
          }
```

datasets/titanic/metadata.json
pierrot0: @benjelloun reminded me that the spec says " In addition to simple atomic types, we also support semantic types, ..." for `dataType` (https://docs.google.com/document/d/1EDUbwwSeSUpRZWczy6FNuCXTbKh9e5H4xXMnqgoCRj8/edit#bookmark=id.o1crjfray9s2).

schema.org definitions are both semantic (eg: MusicAlbum) and not (eg: Integer), so it might feel strange to have two different fields here. 

On the other hand, I can see an advantage to having both a `semanticType` and a `dataType`: it would make it easier to ignore one depending on the use case: when loading the data to tensors for training, `semanticType` can be ignored, while it's likely more useful when doing data visualization.

If `dataType` can contain both a data type definition and a semantic type definition, then we'll need some heuristic to distinguish one from the other.

datasets/titanic/metadata.json
josvandervelde: Agree! That's better

datasets/titanic/metadata.json
pierrot0: Here `references` is the [existing mechanism](https://docs.google.com/document/d/1EDUbwwSeSUpRZWczy6FNuCXTbKh9e5H4xXMnqgoCRj8/edit#bookmark=id.hz571yc5evxj) to join two record sets, and here we are re-using this mechanism.

Using the joining mechanism "as it" has two implications I can think of:
  1. nothing in the current spec says the two `RecordSet`s have to define the same set of keys (there could be more keys defined in `ports_enums`, or less). This is not necessarily a good thing in that case, and to load the data, it would be good to know upfront the definitive number of enum values, or at least a close upper bound.
  2. The "join semantic" is not obvious: are we looking at two different genuine `RecordSet`s (like "movies" and "ratings" in [MovieLens example](https://docs.google.com/document/d/1EDUbwwSeSUpRZWczy6FNuCXTbKh9e5H4xXMnqgoCRj8/edit#heading=h.27t88xicctl), or are we defining the set of possible values?

So indeed, I agree that it might be opportune to create `possible_values` field, or to specify a "join semantic" with references.

@benjelloun what do you think?

datasets/titanic/metadata.json
josvandervelde: Excellent point. I completely agree with you on a conceptual level. 

On a practical level, I'm having some doubts on the complexity of this representation. We're re-using an existing mechanism (recordSet). If we were to use a dedicated mechanism for possible values, with "key", "label", and "url" (maybe "description"?) pre-defined, it would be easier to understand just by glancing at the `json`.

Just as a discussion point, I'm interested in your thoughts!

datasets/titanic/metadata.json
benjelloun: I considered allowing references to RecordSets as an allowed value for dataType... But I stopped short of doing that, because that would introduce RecordSets as a type definition mechanism. 

I think it's better to keep types and possible values separate, and "references" is essentially a foreign key (so, possible values) mechanism.

On your specific points:
Re: 1. Yes, that's correct, and I don't think that's a problem. The target of "references" defines the set of possible values, and the source may not use all of those. The fraction of values that are actually used is data-dependent, and could be captured by computed statistics, but I don't think it should be represented in the schema.
Re: 2. I think the reference in MovieLens has the same semantics as this one: movies in the Ratings table exist in the Movies table. The actual join / flattening is done elsewhere, by creating a new Movies+Ratings RecordSet that combines Movies and Ratings. The reference information is just useful to know how to do the joining.

So for me "references" really means "possible_values", aka, foreign key. I don't see why we would need two separate constructs here. Or am I missing something?

datasets/titanic/metadata.json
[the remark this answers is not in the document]
benjelloun: Great question! I started out using @id, but then switched to name, because the semantics of id doesn't fully match what we're doing, as it's more about defining absolute URLs, and using fragments to abbreviate them rather than defining and referencing local identifiers. CSVW made the same choice.

We could revisit using @id if we think we can make it work. But things like referencing fields of RecordSets with a syntax like "#{rs1/field2}" are a bit of a stretch.

datasets/titanic/metadata.json
benjelloun: Re heuristic: I think we can document the set of atomic data types we support, and require that each field specifies at least one of them. In addition to that, we may publish a set of recommended semantic types / vocabularies, esp. for data that is commonly used in ML datasets.

benjelloun (approved):

datasets/titanic/metadata.json
josvandervelde: That's a good point, I think, but I'm still having some doubts whether we really need this expressiveness.

For any enum, there will be a single field in the data, if I'm not mistaken. E.g. `capital=0` or `capital=1`. If `key=0` then  `label="France, Paris"`, if `key=1` then `label="Netherlands, Amsterdam"`. It's true that you lose some semantic information by just storing the enum's possibilities in a single label, but I'm not sure if this semantic information is worth the additional complexity.

Side thought: could we also map multiple fields to a single `RecordSet`? E.g. `cities` will have different possible values based on the `country` field? I'm not sure if we can handles these cases now (do we want to?)

In any case, I'm just playing the devils advocate here, I'm happy to go along with this solution.

datasets/titanic/metadata.json
josvandervelde: Thanks for this explanation, I agree!

josvandervelde (approved):

pierrot0:
I'll merge the PR, with no further updates. Then I'll incorporate new constructs into the spec.

PR #10, opened 2023-04-21
Add wd:82799 dataType to `label` field (gender_enums, ports_enums) on Titanic dataset.

pierrot0:
https://www.wikidata.org/wiki/Q82799

This would be to indicate that the string is a label mainly targeting human usage.


benjelloun (commented):
That's an interesting one!

names being very common, I'm not sure we want to use wikidata for typing. Since we're primarily based on schema.org, then sc:name (which is equivalent to wd:82799) would be a natural candidate.

The part that is a bit weird is that schema.org/name is a property, not a class. We could either say that we allow both classes and properties for semantic typing (that would be my preference), or add a separate field to model properties.

josvandervelde:
I agree with Omar's comment, no further comments from me :)

pierrot0:
"sc:name" SGTM, code updated.

josvandervelde (approved):

PR #11, opened 2023-04-24
Add basic validator.

marcenacp:
This is the first version. For the moment, I prototyped using
`datasets_format/datasets/titanic/metadata.json`.

Design elements:

- The entry is the script `datasets_format/python/format/scripts/validate.py`.
- README in `datasets_format/python/format/README.md`.
- `Dataset` represents a graph from a file.
- In `Dataset.__init__`, we call the validator.
- The validator builds an RDFLib graph, then converts it to a NetworkX
  graph, then checks this graph.
- At the moment, we only check that each subject has the proper
  properties.
- I put in place some unit tests, particularly in `datasets_format/python/format/src/graphs_test.py`.

Questions:

- What should I expect on `http://mlcommons.org/schema/dataType`? Just that the dataType is a valid URI? Should we fetch the URI?

Todo:

- Check the ID, and run a BFS to check that the tree is valid.
- Launch tests in CI/CD.
- Support more complex use cases (e.g., `ml:RecordSet` -> `data` in Titanic), and other datasets.
- Make OpenML converter compatible (for instance, at the moment, I have the following errors):

```bash
format.src.errors.ValidationError: Found the following 6 error(s) during the validation:
  -  [dataset(Titanic)] Property "https://schema.org/citation" is mandatory, but does not exist. Existing properties: {rdflib.term.URIRef('https://schema.org/version'), rdflib.term.URIRef('https://schema.org/keywords'), rdflib.term.URIRef('https://schema.org/distribution'), rdflib.term.URIRef('https://schema.org/dateModified'), rdflib.term.URIRef('https://schema.org/description'), rdflib.term.URIRef('https://schema.org/url'), rdflib.term.URIRef('https://schema.org/creativeWorkStatus'), rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type'), rdflib.term.URIRef('https://schema.org/@language'), rdflib.term.URIRef('https://schema.org/recordSet'), rdflib.term.URIRef('https://schema.org/isAccessibleForFree'), rdflib.term.URIRef('https://schema.org/name'), rdflib.term.URIRef('https://schema.org/dateCreated')}.
  -  [dataset(Titanic)] Property "https://schema.org/license" is mandatory, but does not exist. Existing properties: {rdflib.term.URIRef('https://schema.org/version'), rdflib.term.URIRef('https://schema.org/keywords'), rdflib.term.URIRef('https://schema.org/distribution'), rdflib.term.URIRef('https://schema.org/dateModified'), rdflib.term.URIRef('https://schema.org/description'), rdflib.term.URIRef('https://schema.org/url'), rdflib.term.URIRef('https://schema.org/creativeWorkStatus'), rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type'), rdflib.term.URIRef('https://schema.org/@language'), rdflib.term.URIRef('https://schema.org/recordSet'), rdflib.term.URIRef('https://schema.org/isAccessibleForFree'), rdflib.term.URIRef('https://schema.org/name'), rdflib.term.URIRef('https://schema.org/dateCreated')}.
  -  [dataset(Titanic) > distribution(Titanic (arff))] Property "https://schema.org/contentSize" is mandatory, but does not exist. Existing properties: {rdflib.term.URIRef('https://schema.org/name'), rdflib.term.URIRef('https://schema.org/encodingFormat'), rdflib.term.URIRef('https://schema.org/contentUrl'), rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type')}.
  -  [dataset(Titanic) > distribution(Titanic (arff))] At least one of these properties should be defined: [rdflib.term.URIRef('https://schema.org/sha256'), rdflib.term.URIRef('https://schema.org/md5')]. Existing properties: {rdflib.term.URIRef('https://schema.org/name'), rdflib.term.URIRef('https://schema.org/encodingFormat'), rdflib.term.URIRef('https://schema.org/contentUrl'), rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type')}.
  -  [dataset(Titanic) > distribution(Titanic (parquet))] Property "https://schema.org/contentSize" is mandatory, but does not exist. Existing properties: {rdflib.term.URIRef('https://schema.org/name'), rdflib.term.URIRef('https://schema.org/encodingFormat'), rdflib.term.URIRef('https://schema.org/contentUrl'), rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type')}.
  -  [dataset(Titanic) > distribution(Titanic (parquet))] At least one of these properties should be defined: [rdflib.term.URIRef('https://schema.org/sha256'), rdflib.term.URIRef('https://schema.org/md5')]. Existing properties: {rdflib.term.URIRef('https://schema.org/name'), rdflib.term.URIRef('https://schema.org/encodingFormat'), rdflib.term.URIRef('https://schema.org/contentUrl'), rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type')}.
```


datasets/titanic/metadata.json
```diff
@@ -32,11 +45,11 @@
         "key": "#{key}",
         "field": [
             { "name": "key", "@type": "ml:Field", "dataType": "sc:Integer" },
-            { "name": "label", "@type": "ml:Field", "dataType": "sc:String" },
+            { "name": "label", "@type": "ml:Field", "dataType": "sc:Text" },
             {
                 "name": "url",
                 "@type": "ml:Field",
-                "dataType": ["sc:Url", "wd:48277"]
+                "dataType": ["sc:url", "wd:48277"]
```
benjelloun, line 52: sc:url is a property, not a type. I think you want sc:URL here.

We discussed with @pierrot0  whether we want to allow properties as well, but I don't think we reached a definitive conclusion on that yet.

python/format/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[project]
+name = "format"
```
benjelloun, line 2: Should this be called sth like validator instead?

python/format/src/datasets.py
```diff
@@ -0,0 +1,55 @@
...
+def _load_file(file: FileOrFilePath) -> dict:
+    if isinstance(file, str):
+        file = epath.Path(file)
+        if not file.exists():
+            raise ValueError(f"File {file} does not exist.")
+    if isinstance(file, epath.PathLike):
+        with file.open() as file:
+            file = json.load(file)
+    if not isinstance(file, dict):
+        raise ValueError("The file is not a valid JSON-LD, because it's not a dict.")
```
benjelloun, line 23: Nit: Being a "dict" sounds like a Python specific attribute. In JSON terminology, should that be an "object" instead?

python/format/src/errors.py
```diff
@@ -0,0 +1,62 @@
...
+class ValidationError(Exception):
+    """Error during the validation of the format."""
+
+
+@dataclasses.dataclass
+class Issues:
+    """
+    Issues during the validation of the format.
+
+    Issues can either errors (blocking) or warnings (informative).
```
benjelloun, line 15: Nit: either *be* errors...

python/format/src/errors.py
```diff
@@ -0,0 +1,62 @@
...
+    def report(self) -> str:
+        """Reports errors and warnings in a string."""
+        message = ""
+        for issues, issue_type in [
+            (self.errors, "error(s)"),
+            (self.warnings, "warning(s)"),
+        ]:
+            num_issues = len(issues)
+            if num_issues:
+                message += f"\nFound the following {len(issues)} {issue_type} during the validation:"
```
benjelloun, line 44: Curious why the \n is at the beginning, not at the end of the messages.

benjelloun (changes requested):
Looks good! 
One general question: Do we want so many test files, given the syntax may still change? Is there a way to create fewer of them, and still test all the cases we want to cover?

datasets/titanic/metadata.json
marcenacp: What should the verifier check for as far as dataTypes are concerned? How can I verify that a URI defines a dataType?

python/format/pyproject.toml
marcenacp: Seen together: for the moment we keep the current name, because data verification and data generation are very entangled, and it's not clear at this stage how we will split them in two libraries.

python/format/src/errors.py
marcenacp: Done.

marcenacp:
@benjelloun, thank you for the review!

I also addressed your question on reducing the number of test files. I think we have to keep a minimum of "end-to-end tests" with real data, but I added more unit tests to support other cases.

marcenacp:
@pierrot0 Can you please review? Notably, I would like your input on the project's structure:

```
.
|-- python
|   |-- format
|   |   |-- __init__.py
|   |   |-- pyproject.toml
|   |   |-- README.md
|   |   |-- scripts
|   |   |   `-- validate.py
|   |   `-- src
|   |       |-- constants.py
|   |       |-- datasets.py
|   |       |-- datasets_test.py
|   |       |-- errors.py
|   |       |-- errors_test.py
|   |       |-- graphs.py
|   |       |-- graphs_test.py
|   |       |-- __init__.py
|   |       `-- tests
|   |           `-- graphs
|   |               ... *.json files for tests
|   `-- openmlconverter
|       ...
`-- README.md
```

I think:
- We should drop the `src` folder.
- We should have one `datasets_format` library instead of sub-folders (`python/format`).

datasets/titanic/metadata.json
```diff
@@ -1,11 +1,23 @@
...
+        "@type": "@id"
+      },
+      "recordSet": {
+        "@id": "ml:RecordSet",
+        "@type": "@id"
+      }
     },
-
-    "@type": "sc:Dataset",
+    "@type": "Dataset",
```
pierrot0, line 20: Is there a reason to drop the "sc:" prefix?
This was done to make it more obvious to spot what is coming from where (here schema.org).
If we can keep the prefix, I think it makes the config more readable.

datasets/titanic/metadata.json
```diff
@@ -31,10 +43,21 @@
...
-            { "name": "label", "@type": "ml:Field", "dataType": "sc:String" },
+            {
+              "name": "key",
+              "description": "0 or 1",
+              "@type": "ml:Field",
+              "dataType": "sc:Integer"
+            },
+            {
+              "name": "label",
+              "description": "Male or female",
```
pierrot0, line 54: Female (uppercase F)

datasets/titanic/metadata.json
```diff
@@ -31,10 +43,21 @@
         "description": "Maps gender keys (0, 1) to labeled values.",
         "key": "#{key}",
         "field": [
-            { "name": "key", "@type": "ml:Field", "dataType": "sc:Integer" },
-            { "name": "label", "@type": "ml:Field", "dataType": "sc:String" },
+            {
+              "name": "key",
+              "description": "0 or 1",
```
pierrot0, line 48: Did you add the description because it was raising an error or a warning?

datasets/titanic/metadata.json
```diff
@@ -50,10 +73,21 @@
         "description": "Maps Embarkation port initial to labeled values.",
         "key": "#{key}",
         "field": [
-            { "name": "key", "@type": "ml:Field", "dataType": "sc:Text" },
-            { "name": "label", "@type": "ml:Field", "dataType": "sc:Text" },
+            {
+              "name": "key",
+              "description": "Key in the dataset",
```
pierrot0, line 78: To be coherent with the description above, should it be "C, Q, S or ?"?
I am wondering whether we want to recommend adding a description to this "mapping" RecordSets, which are here to map enumeration values to a semantic meaning and human label.

python/format/README.md
```diff
@@ -0,0 +1,22 @@
...
+```bash
+python -m pip install .
+```
+
+## Validate a file
+
+```bash
+python scripts/validate.py --file ../../datasets/titanic/metadata.json
+```
+
```
pierrot0, line 16: Maybe add a sentence on what the behaviour is when the config is valid or invalid: return code is 0 if no errors, otherwise return code is 1 and the first error is displayed, or something alike.

python/format/src/datasets_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+# End-to-end tests on real data. The data is in `tests/end-to-end/*.json`.
+@pytest.mark.parametrize(
+    ["filename", "error"],
+    [
+        # Metadata.
+        [
+            "metadata_missing_property_name.json",
+            'Property "https://schema.org/name" is mandatory, but does not exist.',
```
pierrot0, line 14: I think the test should check the context too, or there is a chance that one uses mlfield_missing_property_name.json as an input and get a passing test.

python/format/src/datasets_test.py
```diff
@@ -0,0 +1,52 @@
+from etils import epath
+from format.src import datasets
+from format.src import errors
+import pytest
+
+
+# End-to-end tests on real data. The data is in `tests/end-to-end/*.json`.
+@pytest.mark.parametrize(
```
pierrot0, line 8: Another way to write this test might be to write the output of the validator against the examples to "golden files". The test would then just check whether produced output matches expected output for all examples. 

Pro:
 - context is the expected one.
 - this provides nice examples of what the validator produces, which can then be embedded in the doc.
 - Easier for reviewers to check changes.

Con:
 - we would need a script to update the golden files, and run such a script every time the output format changes. But I think that's doable (I have done something similar in a previous project).

pierrot0: Also, it's trivial to add examples / tests.

python/format/src/graphs.py
```diff
@@ -0,0 +1,243 @@
...
+    if len(sources) != 1:
+        issues.add_error(f"Trying to define more than one dataset in the file.")
+    return sources[0]
+
+
+def check_metadata(
+    issues: errors.Issues,
+    edges: Edges,
+    dataset_name: str,
+):
```
pierrot0, line 126: Can you add a docstring to those functions please?

pierrot0:
> I think:
> 
> * We should drop the `src` folder.

What would be the gain of dropping the `src` folder?
It feels to me that it might be beneficiary to have a similar structure for all sub-projects (openmlconverter, ...)?

> * We should have one `datasets_format` library instead of sub-folders (`python/format`).

Do you mean that we may want to have multiple github repos?
Or something else?

datasets/titanic/metadata.json
marcenacp: Done.

datasets/titanic/metadata.json
marcenacp: Not specifying a description currently raises a warning.

datasets/titanic/metadata.json
marcenacp: Done.

datasets/titanic/metadata.json
marcenacp: Done.

python/format/README.md
marcenacp: Done.

python/format/src/datasets_test.py
marcenacp: My experience in web development with "snapshot files" is that they are very noisy to update. Particularly, at a time where the format API is likely to change very often, I thought the current solution would be a trade-off between noise/validation.

WDYT of starting with the current solution and update unit tests later if we feel the need?

python/format/src/graphs.py
marcenacp: Done.

python/format/src/datasets_test.py
pierrot0: SGTM

pierrot0 (approved):
LGTM

datasets/titanic/metadata.json
benjelloun: I think for schema.org and other base types we want to support , it's reasonable to check that the type is one that we know about. It could generate a warning in case of an unknown type.

benjelloun (approved):

PR #12, opened 2023-04-25
PASS and MovieLens example datasets

benjelloun:


datasets/pass/data.json
```diff
@@ -0,0 +1,171 @@
...
+              "@type": "ml:Field",
+              "source": "#{metadata/longitude}"
+            }
+          ]
+        },
+        {
+          "name": "date_taken",
+          "@type": "ml:Field",
+          "description": "The date the photo was taken.",
+          "dataType": "Date",
```
josvandervelde, line 165: Should we (have the capability to) specify the DateTime format of this String? And if so, how?
This is an example of the values in this file: `2004-05-17 00:44:29.0`

datasets/pass/data.json
```diff
@@ -0,0 +1,171 @@
+{
+  "@context": {
+    "@vocab" : "https://schema.org/",
+    "ml": "http://mlcommons.org/schema/",
+    "@base": "http://tensorflow.org/pass"
```
josvandervelde, line 5: What is the purpose of this `@base`? I found that `@base` is "Used to set the base IRI against which to resolve those relative IRI references which are otherwise interpreted relative to the document." 
I don't think there are any relative IRI references in this document (maybe I missed some)

datasets/movielens/data.json
```diff
@@ -0,0 +1,244 @@
+{
+  "@context": {
+    "@vocab" : "https://schema.org/",
```
josvandervelde, line 3: @pierrot0 changed the titanic example from using `@vocab` to explicitly using `"sc": "https://schema.org/",`
in https://github.com/mlcommons/datasets_format/commit/35472310fd323549ad3ad574807ed09c20efd509.
Both methods are fine of course, but I like the explicit approach

datasets/movielens/data.json
```diff
@@ -0,0 +1,244 @@
...
+          "name": "movie_id",
+          "@type": "ml:Field",
+          "dataType": "Integer",
+          "source": "#{movies-table/movieId}"
+        },
+        {
+          "name": "title",
+          "@type": "ml:Field",
+          "description": "The title of the  movie with the release year in parentheses.",
+          "dataType": "Text",
```
josvandervelde, line 79: I'm not very familiar with csvw, is the distinction between `"dataType": "Text"` and `"dataType": "String"` deliberate?
They are both used in these datasets.

datasets/movielens/data.json
```diff
@@ -0,0 +1,244 @@
...
+          "name": "rating",
+          "@type": "ml:Field",
+          "description": "The score of the rating on a five-star scale.",
+          "dataType": "Float",
+          "source": "#{ratings-table/rating}"
+        },
+        {
+          "name": "timestamp",
+          "@type": "ml:Field",
+          "dataType": "Integer",
```
josvandervelde, line 125: Do we want all our examples to use semantic datatypes from WikiData, as it's used in https://github.com/mlcommons/datasets_format/blob/main/datasets/titanic/metadata.json? I'm not sure if we should, I'm curious what you think.

If we want semantic datatypes in this document, we could add `wd:186885` to this `dataType` to signify it's a timestamp, and probably add info to the other fields as well.

datasets/movielens/data.json
```diff
@@ -0,0 +1,244 @@
...
+      "@type": "FileObject",
+      "containedIn": "ml-25m-archive",
+      "contentUrl": "genome-tags.csv",
+      "encodingFormat": "text/csv"
+    }
+  ],
+  "recordSet": [
+    {
+      "name": "movies",
+      "@type": "ml:RecordSet",
```
josvandervelde, line 65: There is an open question in the DCF spec document, "Is this type pending in schema.org?" Would this be a good moment to ask Dan Brickley again?

josvandervelde (commented):
Looks good! I like the usage of `FileSet`.
I added some minor comments / points to discuss.

datasets/movielens/data.json
benjelloun: Done. I think we still need the @vocab declaration so that we don't have to prefix all schema.org names (e.g, name, url, description, etc.)

datasets/movielens/data.json
benjelloun: Ack. I will reach out to danbri@ about this.

datasets/pass/data.json
benjelloun: Originally, I wanted to use @id and the relative IRI mechanism for all the identifiers in the DCF file, but we ended up using name instead.

benjelloun (commented):
Some of your comments disappeared when I amended my commit (rookie mistake). I addressed them in this version, and will answer the comments individually.

datasets/movielens/data.json
benjelloun: I don' t think we need semantic annotations for basic data types.

datasets/pass/data.json
benjelloun: Yes! I added a format field, which we should describe in the spec. We could nest it under applyTransform, but I kinda like keeping them separate.

datasets/pass/data.json
```diff
@@ -0,0 +1,171 @@
...
+      "name": "pass0",
+      "@type": "FileObject",
+      "contentUrl": "https://zenodo.org/record/6615455/files/PASS.0.tar",
+      "sha256": "0be3a104d6257d83296460b419f82c71",
+      "encodingFormat": "application/x-tar"
+    },
+    {
+      "name": "pass1",
+      "@type": "FileObject",
+      "contentUrl": "https://zenodo.org/record/6615455/files/PASS.2.tar",
```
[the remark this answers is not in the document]
benjelloun: Fixed.

datasets/pass/data.json
josvandervelde: Ok! Should `@base` then be deleted?

pierrot0:
Thank you Omar, LGTM!

pierrot0:
Thank you Omar, LGTM!

PR #17, opened 2023-05-09
[1/2] Refactor `assert` functions in class

marcenacp:
In the close future, we'll have to write smarter assert functions with finer rules.

Example: this rule was not trivial to express with the previous representation:
- If a distribution is a FileObject, then it should define `contentUrl` and `encodingFormat`.
- Else if a distribution is a FileSet, then it should define `includes` and `encodingFormat`.

The PR allows to write such rules more elegantly in a more concise way.


python/format/src/graphs.py
```diff
@@ -11,6 +14,94 @@
 Edges = [Any, Any, Any]
 
 
+@dataclasses.dataclass(frozen=True)
+class Node:
```
pierrot0, line 18: Might be useful to have a small docstring and give a few examples of nodes.

pierrot0 (dismissed):

python/format/src/graphs.py
marcenacp: Done.

marcenacp:
Thanks for the review, @pierrot0!

PR #20, opened 2023-05-15
Do not raise errors in the validation script

marcenacp:
Only print them to `logging.error`.


python/format/scripts/validate.py
```diff
@@ -14,11 +17,16 @@
...
 
+
 def main(argv):
     del argv
     file = FLAGS.file
-    datasets.Dataset(file)
-    print("Done.")
+    try:
+        datasets.Dataset(file)
+        print("Done.")
```
josvandervelde, line 26: Minor point: wouldn't `logging.info` be better here?

marcenacp: Done.

marcenacp (commented):
Thanks!

PR #21, opened 2023-05-15
Feature/make openml converter pass validation

josvandervelde:
Updated OpenML converter so that it:
- passes the validation tool of Pierre M.
- reflects changes in the handcrafted examples made by Omar

Updated the OpenML birds&titanic jsons. With previous PR (https://github.com/mlcommons/datasets_format/pull/3) I forgot to update the jsons after implementing the PR changes, so you see some changes now, such as a changed order of the fields.

Updated the `ci.yaml` so that every dataset json is checked by the validator


datasets/titanic/titanic_by_openml_converter.json
```diff
@@ -1,124 +1,121 @@
...
...ing the Encyclopedia Titanica and created the dataset here. Some duplicate passengers have been dropped, many errors corrected, many missing ages filled in, and new variables created. \n\nFor more information about how this dataset was constructed:\nhttp://biostat.mc.vanderbilt.edu/wiki/pub/Main/DataSets/titanic3info.txt\n\n\n### Attribute information  \n\nThe variables on our extracted dataset are pclass, survived, name, age, embarked, home.dest, room, ticket, boat, and sex. pclass refers to passenger class (1st, 2nd, 3rd), and is a proxy for socio-economic class. Age is in years, and some infants had fractional values. The titanic2 data frame has no missing data and includes records for the crew, but age is dichotomized at adult vs. child. These data were obtained from Robert Dawson, Saint Mary's University, E-mail. The variables are pclass, age, sex, survived. These data frames are useful for demonstrating many of the functions in Hmisc as well as demonstrating binary logistic regression analysis using the Design library. For more details and references see Simonoff, Jeffrey S (1997): The \"unusual episode\" and a second statistics course. J Statistics Education, Vol. 5 No. 1.",
```
benjelloun, line 10: Are "author", "source", etc. part of the original description in markdown? Or do you generate that mark-down from your metadata? If it's the latter, then it would make sense to output the corresponding schema.org fields instead. If the former, I think it's fine to leave as is. (We don't want to start parsing mark-down to extract these fields.)

datasets/titanic/titanic_by_openml_converter.json
```diff
@@ -1,124 +1,121 @@
...
-      "name": "Titanic (arff)"
+      "md5": "60ac7205eee0ba5045c90b3bba95b1c4"
     }
   ],
-  "isAccessibleForFree": true,
-  "keywords": "text_data",
-  "name": "Titanic",
   "recordSet": [
     {
+      "name": "Titanic",
```
benjelloun, line 30: We should probably enforce lowercase-only characters for names.

datasets/titanic/titanic_by_openml_converter.json
```diff
@@ -1,124 +1,121 @@
...
         {
+          "name": "body",
           "@type": "ml:Field",
           "dataType": "Number",
-          "name": "body",
-          "source": "#{Titanic (parquet)/body}"
+          "source": "#{Titanic_arff/body}"
         },
         {
+          "name": "home.dest",
```
benjelloun, line 113: Nit: Why a "."  in the name?

python/openmlconverter/src/converter.py
```diff
@@ -48,7 +52,7 @@ def convert(openml_dataset: dict, openml_features: list[dict]) -> dict:
         "datePublished": _ds(field="collection_date", transform=_lenient_date_parser),
         "inLanguage": _ds(field="language", transform=lambda v: langcodes.find(v).language),
         "isAccessibleForFree": True,
-        "license": _ds(field="license"),
+        "license": _ds(field="licence"),
```
benjelloun, line 55: Typo on "license"

datasets/titanic/titanic_by_openml_converter.json
josvandervelde: Unfortunately, this is part of the original description. The person who created this dataset had the option to put the author in the `author` field, but decided not to.

python/openmlconverter/src/converter.py
josvandervelde: Not a typo, OpenML uses `licence` instead of `license`. I just read up on it, it is UK English.

datasets/titanic/titanic_by_openml_converter.json
josvandervelde: As name I used the exact name of this attribute/column: `@attribute 'home.dest' string`
Should I clean up these values (remove capitals and special characters)? 
If so, where do we put the literal name of the attribute/column?

datasets/titanic/titanic_by_openml_converter.json
josvandervelde: Fixed in https://github.com/mlcommons/datasets_format/pull/21/commits/446b14d81d70110fdf9f88b7baef2fd148958957

josvandervelde:
@benjelloun @PGijsbers 
Many thanks for the reviews! It took some more time than I hoped, but I think I fixed the issues. Could you take a second look?

datasets/titanic/titanic_by_openml_converter.json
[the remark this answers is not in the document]
josvandervelde: Excellent point, fixed it here: feature/make_openml_converter_pass_validation

datasets/titanic/titanic_by_openml_converter.json
```diff
@@ -1,124 +1,121 @@
 {
   "@context": {
     "@vocab": "https://schema.org/",
+    "sc": "https://schema.org/",
     "ml": "http://mlcommons.org/schema/"
```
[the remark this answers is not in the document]
josvandervelde: Fixed in https://github.com/mlcommons/datasets_format/pull/21/commits/91de946605cc244b618b2522e69227527d8c2501

benjelloun (approved):

PR #30, opened 2023-05-16
Set up Pylint

marcenacp:


pierrot0 (approved):
Thanks!

.vscode/settings.json
```diff
@@ -0,0 +1,20 @@
+{
+    "[python]": {
+        "editor.defaultFormatter": "ms-python.black-formatter",
+        "editor.rulers": [
```
josvandervelde, line 4: It would be nice to have an IDE-agnostic solution, I think. It's still fine to add `.vscode` integration, but configuration might be better outside of vscode.
Would it be possible to put the pylint config in `.pylintrc`, the black in the `pyproject.toml`, etc.?

josvandervelde (commented):
Great to welcome Pylint and Black!
Should we also setup some project-wide pre-commit hooks? This would help if you don't use vscode.

.vscode/settings.json
marcenacp: Good point! I already included changes in python/format/.pylintrc and python/format/pyproject.toml. In the future, if we want to support other IDEs, we can integrate specific configurations for those.

PR #34, opened 2023-05-19
Minimal recipes

josvandervelde:
First start with the recipes: two minimal example jsons. 

1. A dataset with only the required fields. Passes the validator, but gives warnings
2. A dataset with only the recommended fields. Passes the validator without warnings.

Discussion:
- should `url` be required (as is) or recommended? It's less essential than name/description, I'd say.


datasets/recipes/minimal.json
benjelloun, line 1: I don't know if this example is really needed. It's basically a schema.org/Dataset minimal dataset, which should be covered by the schema.org documentation.

datasets/recipes/minimal_recommended.json
```diff
@@ -0,0 +1,49 @@
...
+    "sc": "https://schema.org/",
+    "ml": "http://mlcommons.org/schema/",
+    "recordSet": "ml:RecordSet",
+    "field": "ml:Field"
+  },
+  "@type": "sc:Dataset",
+  "name": "Minimal example with recommended fields",
+  "description": "This is a minimal example, including the required and the recommended fields.",
+  "url": "https://example.com/dataset/minimal-recommended/about",
+  "citation": "@article{nash51, author = \"Nash, John\", title  = \"Non-cooperative Games\", journal = \"Annals of Mathematics\", year = 1951, volume = \"54\", number = \"2\", pages  = \"286--295\"}",
```
benjelloun, line 13: I don't think we need to add citation into the "recommended" minimal example.

datasets/recipes/minimal_recommended.json
```diff
@@ -0,0 +1,49 @@
+{
+  "@context": {
+    "@vocab": "https://schema.org/",
+    "sc": "https://schema.org/",
+    "ml": "http://mlcommons.org/schema/",
+    "recordSet": "ml:RecordSet",
```
benjelloun, line 6: Down the line, we probably want to shove all these definitions into an external context file, as specified in:
https://www.w3.org/TR/json-ld/#advanced-context-usage

I'm not 100% sure this is supported by Search engines though.

datasets/recipes/minimal_recommended.json
```diff
@@ -0,0 +1,49 @@
...
+      "encodingFormat": "text/csv"
+    }
+  ],
+  "recordSet": [
+    {
+      "name": "examples",
+      "@type": "ml:RecordSet",
+      "source": "#{examples-table}",
+      "key": "#{hash}",
+      "description": "Describing the columns of the table.",
```
benjelloun, line 30: Maybe the description should rather be sth like: "Records extracted from the examples table, with their schema."

datasets/recipes/minimal_recommended.json
```diff
@@ -0,0 +1,49 @@
...
+      "sha256": "0b033707ea49365a5ffdd14615825511",
+      "encodingFormat": "text/csv"
+    }
+  ],
+  "recordSet": [
+    {
+      "name": "examples",
+      "@type": "ml:RecordSet",
+      "source": "#{examples-table}",
+      "key": "#{hash}",
```
benjelloun, line 29: There is no "#{hash}" field in this table :) Maybe just omit the key?

datasets/recipes/minimal_recommended.json
```diff
@@ -0,0 +1,49 @@
...
+  "name": "Minimal example with recommended fields",
+  "description": "This is a minimal example, including the required and the recommended fields.",
+  "url": "https://example.com/dataset/minimal-recommended/about",
+  "citation": "@article{nash51, author = \"Nash, John\", title  = \"Non-cooperative Games\", journal = \"Annals of Mathematics\", year = 1951, volume = \"54\", number = \"2\", pages  = \"286--295\"}",
+  "license": "https://creativecommons.org/licenses/by/4.0/",
+  "distribution": [
+    {
+      "@type": "sc:FileObject",
+      "name": "examples-table",
+      "contentUrl": "https://example.com/dataset/minimal-recommended/data.csv",
```
benjelloun, line 19: Should a sample CSV file also be included in this example?

benjelloun (approved):

datasets/recipes/minimal.json
pierrot0: I think it might be helpful to start the documentation.
Rather than saying "look at schema.org doc for a minimal example, we can show snippets of this one, and from there link to the next examples.

datasets/recipes/minimal_recommended.json
pierrot0: +1

datasets/recipes/minimal_recommended.json
pierrot0: https://github.com/mlcommons/datasets_format/issues/42 to track.

pierrot0 (approved):
Thanks Jos!

datasets/recipes/minimal_recommended.json
josvandervelde: I removed the citation. This does mean that we get a warning in the validator:
`[dataset(Minimal example with recommended fields)] Property "https://schema.org/citation" is recommended, but does not exist.`

Should the validator be updated so that this warning is not raised? @marcenacp

datasets/recipes/minimal_recommended.json
```diff
@@ -0,0 +1,47 @@
+{
+  "@context": {
+    "@vocab": "https://schema.org/",
+    "sc": "https://schema.org/",
+    "ml": "http://mlcommons.org/schema/",
+    "recordSet": "ml:RecordSet",
+    "field": "ml:Field"
```
marcenacp, line 7: Only missing:

```
    "dataType": "ml:dataType",
    "source": "ml:source",
    "references": "ml:references",
```

PR #37, opened 2023-05-23
Recipes: Enum + Untar

josvandervelde:
Two simple recipes based on PASS (Untar) and Titanic (enums).
They contain all required and recommended fields, so they pass validation without warnings

See https://github.com/mlcommons/datasets_format/pull/34 for the minimal recipes.


datasets/recipes/enum.json
```diff
@@ -0,0 +1,76 @@
...
+          { "key": 1, "label": "East" },
+          { "key": 2, "label": "South" },
+          { "key": 3, "label": "West" }
+      ]
+    },
+    {
+      "@type": "ml:RecordSet",
+      "name": "examples",
+      "source": "#{examples-table}",
+      "description": "Descring the columns of the table",
```
PGijsbers, line 56:
```suggestion
      "description": "Describing the columns of the table",
```

datasets/recipes/enum.json
```diff
@@ -0,0 +1,76 @@
...
+    "ml": "http://mlcommons.org/schema/",
+    "includes": "ml:includes",
+    "recordSet": "ml:RecordSet",
+    "field": "ml:Field"
+  },
+  "@type": "sc:Dataset",
+  "name": "Enum example",
+  "description": "This is a fairly minimal example, showing a way to describe enumerations.",
+  "url": "https://example.com/dataset/enum/about",
+  "citation": "@article{nash51, author = \"Nash, John\", title  = \"Non-cooperative Games\", journal = \"Annals of Mathematics\", year = 1951, volume = \"54\", number = \"2\", pages  = \"286--295\"}",
```
pierrot0, line 14: Maybe remove citation and anything which is not strictly necessary and now showcased in the example itself?

datasets/recipes/enum.json
```diff
@@ -0,0 +1,76 @@
+{
```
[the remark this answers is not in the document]
josvandervelde: Definitely! Ideally all `.json` files are checked automatically, but the openml-generated jsons are invalid.
I fixed that in https://github.com/mlcommons/datasets_format/pull/21. Automatic validation is also included in that PR.

Is it OK to wait on that PR to merge, and ignore it in this PR?

datasets/recipes/compressed_archive.json
```diff
@@ -0,0 +1,49 @@
...
+    "recordSet": "ml:RecordSet",
+    "field": "ml:Field"
+  },
+  "@type": "sc:Dataset",
+  "name": "Compressed archive example",
+  "description": "This is a fairly minimal example, showing a way to describe archive files.",
+  "url": "https://example.com/datasets/recipes/compressed_archive/about",
+  "distribution": [
+    {
+      "name": "archive",
```
pierrot0, line 16: I don't know if that is a consensus, but I think we should name the files using the same filename as much as possible, here: "compressed_archive.zip".

This makes it easier to follow later, but also would make it easier for libraries such as TFDS or HF to save the file with the most correct name on disk, especially when filename cannot be inferred from URL. It would also help in cases where resources need to be downloaded manually (there are a few in TFDS for example) because of various reasons (user needs to approve license, etc.).

datasets/recipes/compressed_archive.json
```diff
@@ -0,0 +1,49 @@
...
+      "@type": "sc:FileObject",
+      "contentUrl": "https://example.org/datasets/recipes/compressed_archive.zip",
+      "sha256": "4cee 7cd36f1d874072b6a96c95ce53c9b01d158f0737e754e6c9cc80aeb06e5a",
```
pierrot0, line 19: Remove space within checksum?

pierrot0 (approved):
LGTM, thanks Jos!

datasets/recipes/compressed_archive.json
josvandervelde: I think that's an excellent suggestion! Fixed in https://github.com/mlcommons/datasets_format/pull/37/commits/0553950c6f24146b7f5c9fba88abf92c7d738144

datasets/recipes/enum.json
```diff
@@ -0,0 +1,74 @@
...
+          },
+          {
+            "name": "label",
+            "description": "North, East, South or West",
+            "@type": "ml:Field",
+            "dataType": ["sc:Text", "sc:name"]
+          }
+      ],
+      "data": [
+          { "key": 0, "label": "North" },
```
marcenacp, line 44: This is not valid in schema.org, as schema.org doesn't accept arbitrary JSON. Can you create a separate CSV and reference it?

datasets/recipes/enum.json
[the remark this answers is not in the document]
marcenacp: Ideally, you would first merge https://github.com/mlcommons/datasets_format/pull/21, then rebase to see that the files in the current PR are valid.

I see a few missing element in the context:

```
"dataType": "ml:dataType",
"source": "ml:source",
"references": "ml:references"
```

marcenacp (approved):
Have you tried to generate the data? Using `script/generate.py`. It may fail (for example, because the CSVs are not in `data/`). Otherwise, LGTM! Thanks.

PR #38, opened 2023-05-23
Refactor from_nodes into smaller functions

marcenacp:


python/format/src/computations.py
```diff
@@ -184,8 +207,14 @@ def __call__(self):
 class ReadField(Operation):
     """Reads a field from a Pandas DataFrame and applies transformations."""
 
+    node: Field
+
     def __call__(self, args: pd.Series):
-        _, field = self.node.source.reference
+        assert len(self.node.source.reference) == 2, (
```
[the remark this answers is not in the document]
marcenacp: Thanks for the feedback, Pieter!

There isn't a clear decision if we should fail fast (raise errors) or gather as many issues as possible before failing. So I created: https://github.com/mlcommons/datasets_format/issues/39

python/format/src/computations.py
```diff
@@ -197,6 +226,72 @@ def __call__(self, args):
         return {self.node.name: args}
 
 
+def _find_record_set(graph: nx.MultiDiGraph, node: Node) -> RecordSet:
+    parent_node = graph.nodes[node].get("parent")
```
pierrot0, line 230: 1line docstrings on each function may help readability.

pierrot0 (approved):
Thank you!

python/format/src/computations.py
marcenacp: Done.

marcenacp: (See next PR)

PR #48, opened 2023-06-05
Allow dot (".") in names. 

pierrot0:
For example to be able to name a file "file.zip".

Related to my comment in https://github.com/mlcommons/datasets_format/pull/37#discussion_r1217890269 and failing tests on https://github.com/mlcommons/datasets_format/actions/runs/5176448809/jobs/9325271427?pr=45.

### DIFF
--- a/python/format/src/datasets_test.py
+++ b/python/format/src/datasets_test.py
@@ -74,7 +74,7 @@ import pytest
         ],
         [
             "mlfield_bad_source.json",
-            "Malformed source data: #{THISDOESNOTEXIST/field}.",
+            "Malformed source data: #{THISDOESNOTEXIST#field}.",
         ],
     ],
 )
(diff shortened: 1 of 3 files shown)


josvandervelde (commented):
A suggestion from Piere M: "I think we should accept anything but /, #, { and }"
What do you think?

marcenacp (approved):

pierrot0:
> A suggestion from Piere M: "I think we should accept anything but /, #, { and }". I think that that's a nice suggestion. What do you think?

I agree. Even `#`, `{` and `}` should be acceptable I think, if instead of `#{filename/columname}` we use `filename/columname` for references. There is still the problem of distinguishing between "builtin" properties like `.filename` and a column name, so we will need some more changes in syntax, but that's for another thread...

pierrot0:
I'll merge this PR with just adding dot (".") to supported characters to unblock other PR, and I'll open a conversation for the  rest + reference mechanism.

PR #50, opened 2023-06-05
Actual checks of `dataType` when generating the dataset

marcenacp:


datasets/titanic/metadata.json
```diff
@@ -116,15 +116,15 @@
             "name": "gender",
             "description": "Gender of passenger (0: Male, 1: Female)",
             "@type": "ml:Field",
-            "dataType": "sc:Integer",
+            "dataType": "sc:Text",
```
marcenacp, line 119: Note to the reviewer: for now we haven't implemented the join with the gender enums, so it's a `Text`, not an `Integer`.

python/format/src/datasets.py
```diff
@@ -46,8 +46,9 @@ class Validator:
     def run_static_analysis(self):
         try:
             file_path, self.file = _load_file(self.file_or_file_path)
-            rdf_graph = graphs.load_rdf_graph(self.file)
-            nodes = graphs.check_rdf_graph(self.issues, rdf_graph)
+            rdf_graph, rdf_nx_graph = graphs.load_rdf_graph(self.file)
+            rdf_namespace_manager = rdf_graph.namespace_manager
```
marcenacp, line 50: Note to the reviewer: I'm maybe thinking of using this namespace manager to refactor how I parse the JSON. I could go from JSON to our graph representation without using the intermediary RDF graph.

datasets/titanic/metadata.json
pierrot0: I would prefer we support casting right away, ie: if the read type (string) is compatible with specified type (number), then just casting to specified type. I don't think this needs to wait for join.

 Alternatively, could you add a README.md in directory with a todo so people who read the example aren't confused?

pierrot0 (approved):

datasets/titanic/metadata.json
pierrot0: Sorry my mistake: it's "female" in dataset, so indeed Text.
Ignore my previous comment, thanks!

python/format/src/datasets.py
pierrot0: SGTM

PR #64, opened 2023-06-08
List some supported dataTypes in the spec

pierrot0:


pierrot0:
I'll follow-up with an how-to page on how to specify ML data splits on a different page.

docs/croissant-spec.md
```diff
@@ -263,14 +263,40 @@ corresponding properties must be defined using `subField`.
 
 A field may have more than a single assigned `dataType`, in which case at least
 one must inform about the expected type of data (eg: `sc:Text`), while other
-types inform about the semantic being used.
+types inform about the semantic being used, possibly semantics with ML meaning.
 
 **range**:	[sc:Text](https://schema.org/Text), [sc:URL](https://schema.org/URL)
```
marcenacp, line 268: More data types are listed here: https://github.com/mlcommons/datasets_format/blob/main/python/format/src/data_types.py

docs/croissant-spec.md
```diff
@@ -263,14 +263,40 @@ corresponding properties must be defined using `subField`.
...
 
 
-In the following example, the `url` field is expected to be a URL , which
+Supported types with ML meaning:
+
+| `dataType` | Usage |
+| -------- | ----- |
+| [**`sc:ImageObject`**](https://schema.org/ImageObject) | Describes a field containing the content of an image (pixels). |
+| [**`ml:BoundingBox`**](http://mlcommons.org/schema/BoundingBox) | Describes a bounding box. |
+| [**`sc:name`**](https://schema.org/name) | Describes a field which can be used as a human-friendly label. |
```
marcenacp, line 279: Should this be in `Supported types with no ML meaning`?

marcenacp (approved):
Great addition!

docs/croissant-spec.md
pierrot0: Thanks, added!

docs/croissant-spec.md
pierrot0: Merged everything in "Supported known types", clearer in the end IMO.

marcenacp (approved):

docs/croissant-spec.md
```diff
@@ -263,14 +263,40 @@ corresponding properties must be defined using `subField`.
...
+| `dataType` | Usage |
+| -------- | ----- |
+| [**`sc:Boolean`**](https://schema.org/Boolean) | Describes a boolean. |
+| [**`sc:Date`**](https://schema.org/Date) | Describes a date. |
+| [**`sc:Float`**](https://schema.org/Float) | Describes a float. |
+| [**`sc:Integer`**](https://schema.org/Integer) | Describes an integer. |
+| [**`sc:Text`**](https://schema.org/Text) | Describes a string. |
+| [**`sc:URL`**](https://schema.org/ImageObject) | Describes a URL. |
+| [**`sc:ImageObject`**](https://schema.org/ImageObject) | Describes a field containing the content of an image (pixels). |
+| [**`ml:BoundingBox`**](http://mlcommons.org/schema/BoundingBox) | Describes a bounding box. |
```
benjelloun, line 283: If we define this under the ml namespace, then we also need to describe it as a class, right? 
We may also consider supporting https://schema.org/GeoShape, which has propperties for box, polygon, etc.

benjelloun: There is also [dcat:bbox](https://www.w3.org/TR/vocab-dcat-3/#Property:location_bbox).

docs/croissant-spec.md
```diff
@@ -263,14 +263,40 @@ corresponding properties must be defined using `subField`.
...
+| [**`sc:name`**](https://schema.org/name) | Describes a field which can be used as a human-friendly label. |
+| [**`wd:Q3985153`**](https://www.wikidata.org/wiki/Q3985153) <br/>(**Training, validation and test sets**) | Describes a field used to divide data into multiple sets according to intended usage with regards to models [training](https://mlcommons.org/definitions/training_split), [validation](https://mlcommons.org/definitions/validation_split), [testing](https://mlcommons.org/definitions/test_split), and possibly others. <br/>While any value is acceptable here, it is recommended to associate the usual splits listed above with the linked semantic URL.
```
benjelloun, line 285: I wonder if we should instead declare a "primary" definition of these types under the ml namespace, and support the wd ones as alternatives. WDYT?

docs/croissant-spec.md
```diff
@@ -263,14 +263,40 @@ corresponding properties must be defined using `subField`.
...
+| [**`wd:Q48277`**](https://www.wikidata.org/wiki/Q48277) <br/>(**gender**) | Describes a field which values are indicative of a person gender. This can be used by Ethical AI tools to flag possible gender bias in the data. Values for this field can be associated with specific gender URLs (eg: [**`wd:Q6581097`**](https://www.wikidata.org/wiki/Q6581097), [**`wd:Q6581072`**](https://www.wikidata.org/wiki/Q6581072), etc.) |
```
benjelloun, line 286: I don't think we should include Gender in this list. It doesn't really fit with the other types, and is just one particular human characteristic that may be important to understand biases in the data. Also, the wd type may not be the canonical way to describe gender information.

Instead, how about adding an "extension mechanism" section to say that we will support additional types, and maybe give this one as an example there?

docs/croissant-spec.md
pierrot0: Added class for ml:BoundingBox, with a format property, so updated COCO dataset accordingly.

Please let me know if that makes sense. It feels somehow akward to me to use the same "ml:" namespace for both:
1. classes, which we can use in `@type` property, and 
2. types we can specify in  `dataType`.

Not sure what `dataType` value should be if we have `"@type": "ml:BoundingBox"`.

I am not sure about using `sc:GeoShape` for bounding boxes. It seems very loosely defined to me... Ditto for `dcat:bbox`. But happy to be convinced otherwise.

docs/croissant-spec.md
pierrot0: SGTM, what about creating those definitions under another namespace?

"ml": "http://mlcommons.org/schema/",
"mld": "http://mlcommons.org/definitions/",

docs/croissant-spec.md
pierrot0: Makes sense. done.

pierrot0:
Happy to do a VC to quickly solve issues.

josvandervelde (approved):
I don't have any additions, looks good to me!

docs/croissant-spec.md
```diff
@@ -368,7 +405,7 @@ mechanisms, or use `sc:PropertyValue` as an escape hatch.
 
 ### format
 
-A format used to parse the values coming from a `DataSource`, e.g., for dates or numbers.
+A format used to parse the values coming from a `DataSource` or a `BoundingBox`, e.g., for dates or numbers.
```
benjelloun, line 408: Why single out Bounding Box? It's just a specific data type that  might be used for a data source, right ? Formats can apply to dates, numbers, bounding boxes, shape files, etc.

datasets/coco2014/metadata.json
```diff
@@ -297,13 +297,11 @@
         },
         {
           "name": "bbox",
-          "@type": "ml:Field",
-          "description": "The bounding box around annotated object[s].",
+          "@type": "ml:BoundingBox",
```
benjelloun, line 300: This should be ml:Field, right? BoundingBox is specificed in the dataType.

datasets/coco2014/metadata.json
```diff
@@ -297,13 +297,11 @@
...
+          "format": "XYWH",
```
benjelloun, line 301: I think format should be inside the source, like it was before.

benjelloun (commented):
Minor syntax comments.

datasets/coco2014/metadata.json
pierrot0: done

datasets/coco2014/metadata.json
pierrot0: done.

pierrot0: done

datasets/coco2014/metadata.json
pierrot0: done

benjelloun (approved):

pierrot0:
> Minor syntax comments.

All fixed.

PR #65, opened 2023-06-08
Simple split example

pierrot0:
A minimal example for splits. The example doesn't contain semantic definition and this is on purpose.

### DIFF
--- a/datasets/recipes/simple-split.json
+++ b/datasets/recipes/simple-split.json
@@ -19,7 +19,7 @@
             "name": "books.csv",
             "@type": "sc:FileObject",
             "contentUrl": "data/books.csv",
-            "sha256": "",
+            "sha256": "d35c5a01eecbd7700faf86b4ec838eb65bd6e861633b1e10ca3294d4e58e75c9",
             "encodingFormat": "text/csv"
         }
     ],


datasets/simple-split/metadata.json
```diff
@@ -0,0 +1,58 @@
...
+    "name": "simple-split",
+    "description": "An artificial example dataset defining splits from a CSV column",
+    "url": "https://mlcommons.org",
+    "license": "https://creativecommons.org/licenses/by/4.0/",
+    "distribution": [
+        {
+            "name": "books.csv",
+            "@type": "sc:FileObject",
+            "contentUrl": "data/books.csv",
+            "sha256": "",
```
josvandervelde: Did you omit the checksum on purpose?
This file's sha256 is d35c5a01eecbd7700faf86b4ec838eb65bd6e861633b1e10ca3294d4e58e75c9

josvandervelde (approved):
Looks good!

datasets/simple-split/metadata.json
pierrot0: Thanks! I suppose it was just laziness, I am trying to use github codespace and haven't found a way to easily get the sha256 from there...

PR #66, opened 2023-06-08
Howto specify splits documentation page

pierrot0:

PR #67, opened 2023-06-09
Wiki-text

josvandervelde:
The croissant for the wiki-text dataset.

### DIFF
--- a/datasets/wiki-text/metadata.json
+++ b/datasets/wiki-text/metadata.json
@@ -94,7 +94,7 @@
                 "replace": "\\n/<eos>"
               },
               {
-                "separator": "|"
+                "separator": " "
               }
             ]
           }
--- a/datasets/wiki-text/metadata.json
+++ b/datasets/wiki-text/metadata.json
@@ -3,6 +3,7 @@
     "@vocab": "https://schema.org/",
     "sc": "https://schema.org/",
     "ml": "http://mlcommons.org/schema/",
+    "wd": "https://www.wikidata.org/wiki/",
     "includes": "ml:includes",
     "recordSet": "ml:RecordSet",
     "field": "ml:Field",
--- a/datasets/wiki-text/metadata.json
+++ b/datasets/wiki-text/metadata.json
@@ -1,0 +1,121 @@
+{
+  "@context": {
+    "@vocab": "https://schema.org/",
+    "sc": "https://schema.org/",
+    "ml": "http://mlcommons.org/schema/",
+    "includes": "ml:includes",
+    "recordSet": "ml:RecordSet",
+    "field": "ml:Field",
+    "subField": "ml:SubField",
+    "dataType": "ml:dataType",
+    "source": "ml:source",
+    "data": "ml:data",
+    "applyTransform": "ml:applyTransform",
+    "format": "ml:format",
+    "regex": "ml:regex",
+    "separator": "ml:separator",
+    "references": "ml:references"
+  },
+  "@type": "sc:Dataset",
+  "@language": "en",
+  "name": "WikiText-103",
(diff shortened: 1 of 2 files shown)


datasets/wiki-text/metadata.json
```diff
@@ -0,0 +1,121 @@
...
+    "data": "ml:data",
+    "applyTransform": "ml:applyTransform",
+    "format": "ml:format",
+    "regex": "ml:regex",
+    "separator": "ml:separator",
+    "references": "ml:references"
+  },
+  "@type": "sc:Dataset",
+  "@language": "en",
+  "name": "WikiText-103",
```
josvandervelde, line 21: Note for the reviewers:
WikiText has multiple "versions": WikiText-2 aims to be of a similar size to the PTB while WikiText-103 contains all articles extracted from Wikipedia.

I added Wikitext-103 for now. Should I also add WikiText-2? And if so, how to combine them? I could create a https://schema.org/DataCatalog instead, containing the two Datasets, but I'm afraid I'm needlessly complicating things. 
Thoughts?

datasets/wiki-text/metadata.json
```diff
@@ -0,0 +1,121 @@
...
+          "name": "word",
+          "@type": "ml:Field",
+          "description": "A word.",
+          "dataType": "sc:Text",
+          "repeated": "true",
+          "source": {
+            "data": "#{token-files/content}",
+            "applyTransform": [
+              {
+                "replace": "\\n/<eos>"
```
josvandervelde, line 94: Note for the reviewers:

1. applyTransform doesn't allow for lists in the current specs. Should it? Or can I do this more elegantly?
2. replace is not part of the current [specs](https://docs.google.com/document/d/1EDUbwwSeSUpRZWczy6FNuCXTbKh9e5H4xXMnqgoCRj8/edit#bookmark=id.z90k3xy7xz3q). Should it be?

datasets/wiki-text/metadata.json
```diff
@@ -0,0 +1,121 @@
+{
+  "@context": {
+    "@vocab": "https://schema.org/",
+    "sc": "https://schema.org/",
+    "ml": "http://mlcommons.org/schema/",
```
pierrot0, line 5: You'll need the following too: 

```json
"wd": "https://www.wikidata.org/wiki/",
```

I know I forgot it in COCO dataset, but it should be back in PR #66 :)

datasets/wiki-text/metadata.json
pierrot0: I think just having a `WikiText-103` dataset is fine. We can later on add a `WikiText-2` later on.

I am not sure we want to embark on using `sc:DataCatalog` for now, and if we do, then I'd suggest we keep the definition of such a Catalog separate from the datasets it contains (ie: have each dataset described in a separate croissant file, and the catalog in another config).

This will come up a lot. For example with COCO2014-COCO2020, which has some overlap between various years. I think it still makes sense to consider them to be different datasets and mark lineage somewhere, although I don't know how.

datasets/wiki-text/metadata.json
pierrot0: 1. Yes I think it might make sense to support list, which unlike dicts are sorted.
2. Replace sounds to me like something we would want to support. Although we probably don't want to support arbitrary transformations, this one sounds simple enough and common too.

pierrot0 (approved):
Thank you Jos!
Not everything used here is defined in the spec, but I would be in favor of merging the PR and then adding support to spec / documentation, and if needed amend syntax as for formalize things.

datasets/wiki-text/metadata.json
josvandervelde: Good one, thanks! https://github.com/mlcommons/datasets_format/pull/67/commits/3b31ad472f913f1b987e2c73fd25b4c091172759

PR #68, opened 2023-06-12
Fix Titanic passengers RecordSet to reference gender_enums and ports_enums RecordSets 

pierrot0:
To be consistent with what is done on COCO2014 with the `categories` RecordSet.


josvandervelde (approved):
LGTM
There is a weird error though in `test_generation_titanic`. I'm not sure what's causing it...

datasets/titanic/data/gender_maps.csv
benjelloun, line 1: Why not just name this file 'genders.csv'?

datasets/titanic/data/gender_maps.csv
```diff
@@ -1,3 +1,3 @@
-key,label,url
-0,male,wd:Q6581097
```
benjelloun, line 2: nit: wd:Q6581097 is not an expanded form URL. Maybe call the colum wikidata_id or sth?

benjelloun (approved):

PR #76, opened 2023-06-20
definitions for `ml:` concepts to be used by Croissant.

pierrot0:
So it seems to me that we don't intend to use them as JSON-LD `@type` but more as `dataType` values. So I kind of feel strange to have them in the spec itself, but could be convinced otherwise.


docs/definitions/validation_split.md
```diff
@@ -0,0 +1,7 @@
+# `ml:validation_split`
+
+Other known names: `dev`, `val`.
```
josvandervelde, line 3: Should `validation` be another known name?

docs/definitions/testing_split.md
```diff
@@ -0,0 +1,7 @@
+# `ml:testing_split`
+
+Other known names: `test`.
+
+The testing split is a data set used to provide an unbiased evaluation of a final model fit on the training data set. This data split might be hidden by the dataset author.
```
josvandervelde, line 5: nit: should maybe be "data set author" instead of "dataset author" for consistency?

josvandervelde (approved):
Great! Left some very minor comments

pierrot0:
> Great! Left some very minor comments

Thanks!

PR #96, opened 2023-07-03
Refactor operations

marcenacp:


pierrot0 (approved):

python/ml_croissant/_src/operation_graph/base_operation.py
```diff
@@ -0,0 +1,29 @@
+"""Base operation module."""
+
+import dataclasses
+
+from ml_croissant._src.nodes import Node
+
+
+@dataclasses.dataclass(frozen=True, repr=False)
+class Operation:
```
josvandervelde, line 9: Suggestion: make this class abstract and make `__call__` an abstract method.
This would allow for the "implement abstract methods" function in IDEs when defining a new Operation

python/ml_croissant/_src/operation_graph/operations/extract.py
```diff
@@ -0,0 +1,53 @@
...
+        includes = fnmatch.translate(self.target_node.includes)
+        url = self.node.content_url
+        archive_file = get_download_filepath(url)
+        hashed_url = get_hash(url)
+        extract_dir = EXTRACT_PATH / hashed_url
+        if not extract_dir.exists():
+            with tarfile.open(archive_file) as tar:
+                tar.extractall(extract_dir)
+        logging.info(
+            "Found directory where data is extracted: %s.", os.fspath(extract_dir)
```
josvandervelde, line 41: Nit: maybe "The data has been extracted to: %s" ("Found directory... " sounds a bit vague to me)

python/ml_croissant/_src/operation_graph/base_operation.py
```diff
@@ -0,0 +1,29 @@
...
+
+    `@dataclass(repr=False)` allows having a readable stringified `str(operation)`.
+
+    Args:
+        node: The node attached to the operation for the context.
+        output: The result of the operation when it is executed (as returned by
+            __call__).
+    """
+
+    node: Node
```
josvandervelde, line 23: This `node` is the input of the `__call__`. The output type is now not defined. I think that it would make defining the graphs easier if we would have type hints on the output. That would make it possible to automatically check if an Operation can follow another Operation (I assume here that the output of `__call__` of an operation, might be needed for the next operation. I'm not sure exactly how this is implemented, so I might misunderstand some parts). 
It could be implemented using generics. Maybe this is over-engineering, but I always like it.

python/ml_croissant/_src/datasets.py
josvandervelde, line 1: In general, I like the idea of having separate `Operations`. And of course it has to be defined how the `Operations` will follow eachother. How I understand it, currently, is that:
- In `graph.py` the operations are defined
- In `dataset.py` the operations are called. 

If this is true, I have 1 comment:
I would love it, if it was easier to simply see what Operations will be used. For me, this is quite difficult to see. It would be great if there was something like 
**for fields**
`operations = [If(condition_multiple, Join), ReadField]`
**for FileObjects**
`operations = [If(condition_xtar, Untar), If(condition_csv, ReadCsv)]`
This is just an example of course, but I think that something similar would make it more maintainable.

Anyway, I'd suggest to keep this out of scope, but it might be nice to keep in the back of your head.

josvandervelde (approved):
I like this approach! Added some suggestions

marcenacp:
@josvandervelde Thanks for the comments! I'll address them in another PR if it's okay for you. This PR was just moving files.

PR #99, opened 2023-07-06
coco2014: mark categories RecordSet and supercategory field as Enumeration.

pierrot0:
Again, following https://github.com/mlcommons/croissant/discussions/52.

The `ml:isEnum` is set on the supercategory field, since there are no RecordSet defined for those.

There is some ambiguity here though: which of `name` and `supercategory` is the most numerous?
We don't know, and unless we read the data, with no extra indication of cardinality, there could also be the same cardinality for the two fields.

One possibility: establish the rule that when `"ml:isEnumeration": true` is set on a `RecordSet`, then if that `RecordSet` has a Field with `"ml:isEnumeration": true`, the cardinality of that field must be smaller than the cardinality of the `RecordSet`.  That describes well the use case of categories and super-categories like in coco2014, but I don't know if there are other datasets where this wouldn't hold. If no objections with that, we can go with that rule until we find a counter-example.

### DIFF
--- a/datasets/coco2014/metadata.json
+++ b/datasets/coco2014/metadata.json
@@ -236,7 +236,7 @@
     {
       "name": "categories",
       "@type": "ml:RecordSet",
-      "ml:isEnumeration": "true",
+      "ml:isEnumeration": true,
       "key": "#{id}",
       "field": [
         {
@@ -259,7 +259,7 @@
         {
           "name": "supercategory",
           "@type": "ml:Field",
-          "ml:isEnumeration": "true",
+          "ml:isEnumeration": true,
           "description": "The name of the supercategory.",
           "dataType": [
             "sc:Text",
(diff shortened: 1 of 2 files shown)


datasets/coco2014/metadata.json
```diff
@@ -236,6 +236,7 @@
     {
       "name": "categories",
       "@type": "ml:RecordSet",
+      "ml:isEnumeration": "true",
```
marcenacp, line 239: I think JSON supports `true` (without using a string). Should you use a boolean or a string value?

marcenacp (approved):

datasets/coco2014/metadata.json
pierrot0: Absolutely, thanks!

marcenacp (approved):

PR #106, opened 2023-07-10
Add back movielens and wiki-text to the CI

marcenacp:
Features:

- The verifier now supports `apply_transform` passed as a list (like in wiki-text).
- Fields can check their data type (`field.data_type`) from any parent fields if it's not declared on them directly (like in movielens).

Note: this PR adds back movielens to the "structure" verifier, but not to the "operation" verifier (still an early return).

Fixes: https://github.com/mlcommons/croissant/issues/103
Fixes: https://github.com/mlcommons/croissant/issues/101

### DIFF
--- a/python/ml_croissant/ml_croissant/_src/datasets.py
+++ b/python/ml_croissant/ml_croissant/_src/datasets.py
@@ -18,8 +18,8 @@ from ml_croissant._src.operation_graph.operations import (
 from ml_croissant._src.structure_graph.graph import (
     check_structure_graph,
     from_file_to_json,
-    from_json_to_jsonld,
-    from_jsonld_to_nodes,
+    from_json_to_rdf,
+    from_rdf_to_nodes,
     from_nodes_to_structure_graph,
 )
 import networkx as nx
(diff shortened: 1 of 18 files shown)


python/ml_croissant/ml_croissant/_src/datasets.py
```diff
@@ -36,23 +37,24 @@ class Validator:
...
             )
+            check_structure_graph(self.issues, structure_graph)
             # Draw the structure graph for debugging purposes.
             if debug:
                 graphs_utils.pretty_print_graph(structure_graph, simplify=True)
             # Feature toggling: do not check for MovieLens, because we need more
             # features.
             if entry_node.uid == "Movielens-25M":
-                return
+                pass
```
pierrot0, line 57: Remove if and comment above?

pierrot0 (approved):
Thanks!

python/ml_croissant/ml_croissant/_src/datasets.py
marcenacp: Done.

Thanks!

PR #124, opened 2023-07-13
Croissant to Croissant: all datasets

marcenacp:


datasets/coco2014/metadata.json
```diff
@@ -68,147 +69,138 @@
...
+          }
         }
-      ]
+      ],
+      "key": "#{img_id}"
     },
     {
-      "name": "captions",
       "@type": "ml:RecordSet",
-      "key": "#{id}",
```
pierrot0, line 193: I feel like it would be clearer to have the key above field (field at the bottom). Your call.

pierrot0 (approved):
Thanks!

PR #125, opened 2023-07-13
Properly handle `dataType` by value expansion

marcenacp:
Value expansion: https://www.w3.org/TR/json-ld11-api/#value-expansion

An interesting link to play with JSON-LD: https://json-ld.org/playground


pierrot0 (approved):
Excellent!

PR #130, opened 2023-07-14
PASS: support concatenating multiple FileSets/FileObjects

marcenacp:
Before:
image

After:
image


python/ml_croissant/ml_croissant/_src/structure_graph/graph.py
```diff
@@ -86,15 +86,27 @@ def from_json_to_rdf(data: Json) -> rdflib.Graph:
     return graph
 
 
+def _append_to_list(dict: dict[str, list[Any]], key: str, value: Any):
```
pierrot0, line 89: You may be able to use a defaultdict instead: https://docs.python.org/3/library/collections.html#collections.defaultdict

```py
d = defaultdict(list)
d[key].extend([1,2,3])
```

pierrot0 (approved):

python/ml_croissant/ml_croissant/_src/structure_graph/graph.py
marcenacp: Done.

marcenacp:
Thanks!

PR #132, opened 2023-07-14
Enumerations, labels and classes documentation

pierrot0:
This is an attempt to define labels and classes and describe how to use them, as well as enumerations.


docs/howto/enumerations.md
```diff
@@ -0,0 +1,100 @@
...
+2. Using a dedicated `ml:RecordSet`, holding the enumeration values. Such
+   definition allows enumerations to be fully defined, with a name and a
+   semantic URL.
+
+### `ml:Field` enumeration
+
+In the following example, extracted from the [coco2014 Croissant definition](https://github.com/mlcommons/croissant/blob/main/datasets/coco2014/metadata.json), the field `supercategory` is declared to be
+an enumeration using the `"ml:isEnumeration": true` property. This means that
+there is a finite number of `supercategory` values, and that it is expected
+that many rows will have the same value.
```
josvandervelde, line 29: For me, this text would be more easily readable if the example was a little more elaborated.
Something like "In the instances_val2014.json, for example, coco2014 describes 80 possible categories. Each category has an identifier, a name (e.g.: "person", "bicycle" or "car") and a supercategory (e.g. "person" or "vehicle").

docs/howto/enumerations.md
```diff
@@ -0,0 +1,100 @@
...
+  concept of enumerations, which should make it fast to list all possible
+  enumerations.\
+  or
+2. the `RecordSet` in which the enumeration is defined is known to be small.
+  This is the case in the example above, as the parent `RecordSet` is being
+  used to define another enumeration.
+
+### `ml:RecordSet` enumeration
+
+In the following example, also extracted from the [coco2014 Croissant definition](https://github.com/mlcommons/croissant/blob/main/datasets/coco2014/metadata.json), the `categories` `ml:RecordSet` is declared as being an enumeration:
```
josvandervelde, line 63: A little elaboration would help again, for my mind at least. Something like:

In the following example, also extracted from the coco2014 Croissant definition, each picture contains one or more annotations. Each annotation contains a `category_id`. In the Croissant definition, this `category_id` should be linked to the list of category values that's described in another json file.

docs/howto/labels.md
```diff
@@ -0,0 +1,116 @@
...
+}
+```
+
+In the above example, `supercategory` is a discrete text label, and the various
+supercategories are all classes.
+
+## Identifying tag
+
+The term "label" can also refer to an identifying tag, or "a kind of tag to
+idendify an object".
```
josvandervelde, line 72: typo: idendify

docs/howto/labels.md
```diff
@@ -0,0 +1,116 @@
...
+            "sc:label"
+          ],
+          "isEnumeration": true,
+}
+```
+
+In the above example, `supercategory` is a discrete text label, and the various
+supercategories are all classes.
+
+## Identifying tag
```
josvandervelde, line 69: For me it is counter-intuitive that the "Label" and the "Identifying tag" are described in the `labels.md` file, while we only want to refer to the former as a "label". Would it make sense to move this part to a separate `.md` file?

docs/definitions/class.md
```diff
@@ -0,0 +1,19 @@
...
+
+Classes allow to differentiate (or to classify)
+the data – or a subset of the data – into a **finite** set of options.
+The different possible options an output variable can take are often
+referred to as classes.
+
+**Note**
+Classes and labels, although sometimes used indistinctively, do not share the
+exact same definition. Be sure to check the
+[`ml:label` definition](/definitions/label). A class is the possible value a
```
josvandervelde, line 15: These links do not seem to work in Github in the current branch (maybe they're not meant to?)

See https://github.com/mlcommons/croissant/blob/pierrot0/labels/docs/definitions/class.md
Clicking on `ml:label definition` leads me to the non-existing 
https://github.com/mlcommons/croissant/blob/pierrot0/labels/definitions/label 
Instead of 
https://github.com/mlcommons/croissant/blob/pierrot0/labels/docs/definitions/label.md

josvandervelde (commented):
Thanks Pierre!
It's nice to have two options of defining an enumeration, one for simple and one for more complex use-cases. I like it! Added some minor / textual comments

docs/definitions/class.md
```diff
@@ -0,0 +1,19 @@
...
+`ml:label` may take, but a `ml:label` is not necessarily a class.
```
ccl-core, line 16: Maybe adding a small example here could make this distinction clearer? E.g. (if I get it correct) 'colour' is a label and 'green', 'red', etc are classes?

docs/definitions/label.md
```diff
@@ -0,0 +1,14 @@
+# `ml:label`
+
+In ML, a **label** refers to the value being predicted i.e. the desired
+output of a model.
+
+While often referring to discrete values (e.g.`{"cat", "dog"}`), a label can also refer to a continuous value, for example a predicted temperature or price.
+
+**Note**
+Because labels often refer to discrete values, it may be used to designate an
```
ccl-core, line 9: they instead of it? Or maybe "the term can be used to" instead of using a pronoun?

docs/howto/enumerations.md
```diff
@@ -0,0 +1,100 @@
...
+
+Enumerations are **finite** sets of values, which may be **named** and may
+have an associated **semantic meaning**.
+
+**Note**
+While enumerations may be used to define [classes](/definitions/classes) or [labels](/definitions/labels), they don't mecessarily share the same meaning, check the definitions.
+
+Although often being of `dataType` [`sc:Text`](http://schema.org/Text), enumerations can also be of other `dataType`s (e.g. `sc:Integer`).
+
+For example, the titanic dataset uses an enumeration to specify
```
ccl-core, line 11: *Maybe*, I'd put this example before the note (right after the definition).

If we leave it here, maybe change the "For example, ..." starting -- as the previous paragraph ends talking about non-text enumerations :)

docs/howto/enumerations.md
ccl-core: +1

docs/howto/labels.md
```diff
@@ -0,0 +1,116 @@
...
+    ],
+    ...
+}
+```
+
+In the above example, the `air_temperature_c` field is declared as being a
+`ml:label`, on top of being a number. This means that the dataset has
+originally been collected to train models that could predict the air
+temperature in ºC, given other parameters. Note that this does not prevent
+someone from using that same dataset for other purposes as predicting air
```
ccl-core, line 41: than instead of as?

ccl-core, line 41: I _think_ it is does not prevent anyone, but not sure here

ccl-core, line 41: Does this mean that a field labeled as a label can also be used e.g. as a feature?

ccl-core (commented):
Thank you Pierre!

docs/definitions/class.md
benjelloun, line 1: Nit: I'm not sure we want to create a separate directory with individual files for definitions.... Maybe these should all just be part of a "Glossary" section in the expanded spec. We can refactor later.

docs/definitions/label.md
```diff
@@ -0,0 +1,14 @@
+# `ml:label`
+
+In ML, a **label** refers to the value being predicted i.e. the desired
```
benjelloun, line 3: Nit: , i.e.,

docs/definitions/label.md
```diff
@@ -0,0 +1,14 @@
...
+output of a model.
+
+While often referring to discrete values (e.g.`{"cat", "dog"}`), a label can also refer to a continuous value, for example a predicted temperature or price.
```
benjelloun, line 6: Are you sure you want to say "predicted" temperature or price, for a label?

docs/definitions/label.md
benjelloun, line 9: it -> they

docs/howto/enumerations.md
```diff
@@ -0,0 +1,100 @@
+# How to specify enumerations
+
+Enumerations are **finite** sets of values, which may be **named** and may
+have an associated **semantic meaning**.
+
+**Note**
+While enumerations may be used to define [classes](/definitions/classes) or [labels](/definitions/labels), they don't mecessarily share the same meaning, check the definitions.
```
benjelloun, line 7: What does  "check the definitions" mean?

docs/howto/enumerations.md
benjelloun, line 63: I got really confused by this example, because I expected it to match the one above. It took me a while to understand that this was for "category" and not "supercategory" like the one above.

docs/howto/enumerations.md
```diff
@@ -0,0 +1,100 @@
...
+
+The field is of `dataType` `sc:Text`, but also has the type `sc:name`. This
+means that `supercategory` is expected to be a text, and that this text is
+the name of the enumeration value   ` naming itself (the
+supercategory), makes sense to humans, and could therefore be used as it in
+UIs.
+
+Although being straightforward, this approach of defining an enumeration has several limitations:
+
+1. The only way to know the cardinality of the enumeration (the number of
```
benjelloun, line 41: Even then, do we guarantee that all possible values of the enumerations appear in the column?

docs/howto/enumerations.md
```diff
@@ -0,0 +1,100 @@
...
+https://github.com/mlcommons/croissant/blob/6c8eca2925191577a7101dcabd0a880aca03cae3/datasets/coco2014/metadata.json#L284-L291
+
+When declaring an enumeration using a `ml:RecordSet`, that `ml:RecordSet` is
+expected to have at least one key field (here named `id`), which may be of
+various `dataType`s, but will most often be of type `sc:Integer` or `sc:Text`.
+
+As also described in the [labels howto](/howto/labels), the `ml:RecordSet` may
+also have fields of the following type:
+
+- `sc:name`, which may be the same field as the key field, if the key is of
```
benjelloun, line 79: Again, this is using an sc property as a type... We don't support that at the moment. How about starting a discussion to figure out if we want to support this feature?

docs/definitions/class.md
marcenacp: `](/definitions/label)` -> `]((/docs/definitions/label))`

docs/howto/labels.md
```diff
@@ -0,0 +1,116 @@
...
+This documentation page covers the following topics:
+
+- Defining a [`ml:label`](/definitions/label), including a [class](/definitions/class).
+- Defining a label following the general definition "a kind of tag to
+  idendify an object" or identifying tag.
+
+## Defining a `ml:label`
+
+To define a field as being a [`ml:label`](/definitions/label), one must just
+add `ml:label` to its `dataType`s, for example:
```
benjelloun, line 16: hmmm.. I'm not convinced we should use dataType to represent the fact that a field is a label. We may want something similar to what we did for enumerations, i.e., an isLabel property that can go on a RecordSet or a Field.

docs/definitions/label.md
```diff
@@ -0,0 +1,14 @@
...
+
+**Note**
+Because labels often refer to discrete values, it may be used to designate an
+[enumeration value](/howto/enumerations) or a class ([definition](/definitions/class)). Also, the general definition of a "label" outside of ML is
+"a kind of tag to idendify an object". To avoid ambiguities and confusion,
```
marcenacp, line 11: identify

docs/howto/labels.md
```diff
@@ -0,0 +1,116 @@
...
+
+https://github.com/mlcommons/croissant/blob/0f8a1b408a6d225bf777a7dbddc47133d0eb0ea6/datasets/coco2014/data/splits.csv
+
+We also note that semantic URLs can also be used to define `dataType`s
+themselves, [as described in the Croissant spec](
+https://github.com/mlcommons/croissant/blob/0f8a1b408a6d225bf777a7dbddc47133d0eb0ea6/docs/croissant-spec.md#datatype).
+
+### Scope of the name and url
+
+A `sc:name` or a `sc:URL` describes the parent entity, whether that's a record
```
benjelloun, line 114: The use of term "parent" is confusing here. Maybe containing entity?

docs/howto/labels.md
```diff
@@ -0,0 +1,116 @@
...
+
+In the above example, `supercategory` is a discrete text label, and the various
+supercategories are all classes.
+
+## Identifying tag
+
+The term "label" can also refer to an identifying tag, or "a kind of tag to
+idendify an object".
+
+In croissant, there are two **complementary** ways to identify a record:
```
benjelloun, line 74: How is this different from the standard way name and url are used in schema.org?

docs/definitions/label.md
marcenacp, line 6: Are bounding boxes also labels?

marcenacp (approved):

docs/croissant-spec.md
```diff
@@ -286,7 +286,7 @@ types inform about the semantic being used, possibly semantics with ML meaning.
...
 | [**`sc:name`**](https://schema.org/name) | Describes a field which can be used as a human-friendly label. |
-| [**`wd:Q3985153`**](https://www.wikidata.org/wiki/Q3985153) <br/>(**Training, validation and test sets**) | Describes a field used to divide data into multiple sets according to intended usage with regards to models [training](https://mlcommons.org/definitions/training_split), [validation](https://mlcommons.org/definitions/validation_split), [testing](https://mlcommons.org/definitions/test_split), and possibly others. <br/>While any value is acceptable here, it is recommended to associate the usual splits listed above with the linked semantic URL.
+| [**`mlc:split`**](https://mlcommons.org/definitions/split) | Describes a field used to divide data into multiple sets according to intended usage with regards to models [training](https://mlcommons.org/definitions/training_split), [validation](https://mlcommons.org/definitions/validation_split), [testing](https://mlcommons.org/definitions/test_split), and possibly others. <br/>While any value is acceptable here, it is recommended to associate the usual splits listed above with the linked semantic URL.
```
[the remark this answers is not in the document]
pierrot0: No strong opinions. We use `"ml": "http://mlcommons.org/schema/"`, but do we want the definitions to be in schema, or do we want them to be in a separate folder? Or can we have them in both? Going with `ml` for now.

docs/howto/enumerations.md
pierrot0: So for enumerations defined as a `ml:Field` enumeration, by definition all the possible values of the enumeration appear in the column (of a given partition).

For enumerations defined as a `ml:RecordSet`, it's a good point: we should specify explicitely what are the guarantees, or say explicitely that there are no guarantees.

I would be in favor of having: "the listed values of the RecordSet enumeration must be a superset of what appears in the joined RecordSet". If there is a need later, we can always add support for "non complete" enumeration RecordSet. 
This allows loading library to know the cardinality of the enumeration, which is important for loading datasets.
What do you think?

docs/howto/enumerations.md
pierrot0: I agree, this might be confusing. What about we create another type?
Something like "ml:Name"?

docs/howto/labels.md
pierrot0: Well in schema.org, name and url are properties, so one must name the field `name` and `url`, no?
Here we want to be able to specify that one of the data columns is actually the name, another the url, without necessary forcing the name of those columns. Does this make sense?

docs/howto/enumerations.md
pierrot0: I guess we have the same problem with "sc:url" which we use as a type in this same PR.
Should we have a type name something like "ml:UrlDefinition"?

benjelloun: url is less of a problem, because we can use the existing class sc:URL. 

I couldn't find a class in schema.org that represent the fact that an object is a name. It also doesn't make much sense to do that...

If we want to go down the path of using existing properties from vocabularies, then I think we should specify a dataType (class) at the level of the RecordSet, and then the fields of the RecordSet can correspond to the properties of that class. We are already doing something along these lines for subFields: For instance, when we say that we have a field of type sc:GeoCoordinates, the subFields are latitude and longitude, which correspond to properties of that class.

Having DataType at the level of the RecordSet would also make it easier to define simple (single-field) enums / RecordSets by just specifying a simple type and not declaring fields.

marcenacp:
Closing this PR as we consolidated everything into (the final specs)[https://mlcommons.github.io/croissant/docs/croissant-spec.html].

PR #134, opened 2023-07-17
Add coco2014-mini: a simplified version of COCO for unit tests

marcenacp:


python/ml_croissant/ml_croissant/_src/operation_graph/operations/extract_test.py
```diff
@@ -5,5 +5,5 @@
 
 
 def test_str_representation():
```
ccl-core, line 7: I'm not familiar with the codebase, but does this mean that we are testing only with empty files? Wouldn't it make sense at some point to also test w. dummy non-empty file(set)s?

marcenacp: In this PR, I add datasets/coco2014-mini/data/train2014.zip. So we do test on non empty files.

This test is for the string representation.

python/ml_croissant/ml_croissant/_src/operation_graph/operations/extract.py
```diff
@@ -18,9 +19,28 @@
 import pandas as pd
 
 
+def should_extract(encoding_format: str) -> bool:
```
ccl-core, line 22: NIT: more than should_extract is seems is_supported_encoding or something similar?

python/ml_croissant/ml_croissant/_src/operation_graph/operations/extract.py
```diff
@@ -44,9 +63,14 @@ def __call__(self):
                     included_files.append(epath.Path(basepath) / file)
         # We need to sort `files` to have a deterministic/reproducible order.
         included_files.sort()
+        fullpaths = [
```
ccl-core, line 66: NIT: Given that this is a long list comprehension, you might want to add a comment to increase readibility (e.g. why is "[1:]" needed?)

ccl-core (approved):
Great, thanks Pierre!

python/ml_croissant/ml_croissant/_src/operation_graph/operations/extract.py
marcenacp: Acknowledged.

This function is used to know whether to extract or not a node. So I updated the docstring.

python/ml_croissant/ml_croissant/_src/operation_graph/operations/extract.py
marcenacp: Done.

PR #137, opened 2023-07-17
Capture value from a regex in mlcommons.org/schema/applyTransform

marcenacp:


python/ml_croissant/ml_croissant/_src/structure_graph/nodes/source.py
```diff
@@ -101,3 +102,25 @@ def from_json_ld(cls, issues: Issues, field: Any) -> "Source":
...
+        if match is None:
+            logging.debug(f"Could not match {source_regex} in {value}")
+            return value
+        for group in match.groups():
+            if group is not None:
+                return group
+    return value
+
+
+def apply_transforms_fn(value: str, source: Source | None = None) -> str:
```
ccl-core, line 120: As a general rule, I think it would be good to have a one-line docstring for all functions (even the ones for which the function name is somewhat already "explicative") :)

python/ml_croissant/ml_croissant/_src/structure_graph/nodes/source_test.py
```diff
@@ -65,3 +69,26 @@ def test_source_parses_list(json_ld, expected_source):
...
+            "train1234",
+            Source(apply_transform=(Transform(regex="(train|val)\\d\\d\\d\\d"),)),
+            "train",
+        ],
+        # Non working capturing group
+        [
+            "foo1234",
+            Source(apply_transform=(Transform(regex="(train|val)\\d\\d\\d\\d"),)),
+            "foo1234",
+        ],
```
ccl-core, line 90: NIT: Perhaps it might be good to test for a chain of transforms as well?

ccl-core (approved):
Thanks!

python/ml_croissant/ml_croissant/_src/structure_graph/nodes/source.py
marcenacp: Done.

python/ml_croissant/ml_croissant/_src/structure_graph/nodes/source_test.py
marcenacp: Acknowledged.

For the moment, we do not use this feature.

marcenacp:
Thank you!

PR #139, opened 2023-07-17
Fix types of recordSet/field/subField

marcenacp:
recordSets/fields/subFields used to be converted to a scalar if there was only one element. This change is needed for a future change concerning: https://github.com/mlcommons/croissant/issues/115.


ccl-core (approved):
Thanks!

PR #145, opened 2023-07-19
Differentiate between attributes (ml:recordSet) and types (ml:RecordSet)

marcenacp:
When speaking to @benjelloun, I realized we mixed attributes and types when manipulating RecordSets, etc.

In this PR:
- I fix the usage of: `ml:recordSet`/`ml:RecordSet`, `ml:field`/`ml:Field`, `ml:subField`.
- I auto-generate JSON-LD test files, so that they have the proper `@context`.

### DIFF
--- a/datasets/birds/birds_by_openml_converter.json
+++ b/datasets/birds/birds_by_openml_converter.json
@@ -10,7 +10,7 @@
       "@id": "ml:dataType",
       "@type": "@vocab"
     },
-    "field": "ml:Field",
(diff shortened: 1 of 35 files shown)


marcenacp:
@ccl-core, that's a very valid point! They are all documented, except `Dataset` (with `distribution` and `recordSet`). So I just created: https://github.com/mlcommons/croissant/issues/147.

Thank you!

PR #146, opened 2023-07-19
Better specify how joins work in the format (left join)

marcenacp:
Fixes: https://github.com/mlcommons/croissant/issues/128


ccl-core (approved):
Thanks!

PR #149, opened 2023-07-19
Migration of references

marcenacp:
Updates: https://github.com/mlcommons/croissant/issues/115


python/ml_croissant/scripts/migrations/migrate.py
```diff
@@ -5,11 +5,47 @@
 """
 
 import json
```
ccl-core, line 34: I assume we don't need tests for migrations, don't we?

ccl-core (approved):
Thanks!

python/ml_croissant/scripts/migrations/migrate.py
marcenacp: This script is not meant to be re-used ever, so I would say no.

datasets/birds/birds_by_openml_converter.json
```diff
@@ -54,1681 +59,3075 @@
...
+            "distribution": "birds.arff"
+          }
         },
         {
           "@type": "ml:Field",
           "name": "audio_ssd5",
           "dataType": "sc:Number",
-          "source": "#{birds.arff/audio.ssd5}"
+          "source": {
+            "dataExtraction": {
```
ccl-core, line 112: Much more elegant and easier to read than the previous version!

python/ml_croissant/scripts/migrations/migrate.py
ccl-core: SG!

python/ml_croissant/ml_croissant/_src/structure_graph/nodes/source.py
```diff
@@ -28,6 +31,32 @@ def parse_reference(issues: Issues, source_data: str) -> tuple[str, ...]:
...
+class FileProperty(enum.Enum):
+    """Lists the intrinsic properties of a file that are accessible from Croissant."""
+
+    content = enum.auto()
+    filename = enum.auto()
+    filepath = enum.auto()
+    fullpath = enum.auto()
+
+
+def is_file_property(file_property: str):
```
ccl-core, line 43: Might be nice to add a test for this if time allows?

marcenacp: Thanks! Just did it in another PR: https://github.com/mlcommons/croissant/pull/150/commits/3f1b79824fa42c289e38a44fc192905d87101c01

PR #154, opened 2023-07-20
Rename `inlineData`->`data`

marcenacp:
`inlineData` was an alias created only because `data` was ambiguous with `source.data`. Now that we removed `source.data`, we can use `inlineData` again.

Only review the last commit. All other commits are from previous PRs to facilitate future rebase.


ccl-core (approved):
Thanks!

PR #155, opened 2023-07-20
Read and parse JSONs (for COCO2014)

marcenacp:
In this PR:

- `ReadCsv` becomes `Read` and can also read JSON.
- We enrich coco2014-mini to read a JSON serie.

image

### DIFF
--- a/python/ml_croissant/ml_croissant/_src/operation_graph/graph.py
+++ b/python/ml_croissant/ml_croissant/_src/operation_graph/graph.py
@@ -19,7 +19,8 @@ from ml_croissant._src.operation_graph.operations import (
     GroupRecordSet,
     InitOperation,
     Join,
-    ReadCsv,
+    ParseJson,
+    Read,
     ReadField,
 )
 from ml_croissant._src.operation_graph.operations.extract import should_extract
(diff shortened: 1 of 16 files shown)


python/ml_croissant/ml_croissant/_src/operation_graph/graph.py
```diff
@@ -19,7 +19,8 @@
     GroupRecordSet,
     InitOperation,
     Join,
-    ReadCsv,
+    ParseJson,
+    Read,
```
ccl-core, line 23: Can you remind me why we prefer `read` vs. `readCsv`?

ccl-core (approved):
Thanks, looks great!

Looks like there are stills some CI tests failing, that you will probably need to fix?

python/ml_croissant/ml_croissant/_src/operation_graph/graph.py
marcenacp: Because now it reads both CSV and JSON. Tomorrow, it should read any file (JSONL, Parquet, etc). See also https://github.com/mlcommons/croissant/issues/79.

Issue #14, opened 2023-05-09
Cleaner unit tests in golden files


marcenacp:
In unit tests, compare pairs (input JSON, output in the command line) in golden files.

This also allows to have up-to-date documentation with good/bad examples.

marcenacp:
```
src
  tests
    distribution_bad_contained_in.json << what we currently have
    distribution_bad_contained_in << what we want
      metadata.json
      output.txt
```

Before we were testing that a string in contained in the error.

Now we want the output == output.txt

PR #160, opened 2023-07-26
Tests for: metadata_missing_property_name

ccl-core:


ccl-core:
This refers to Issue #14 - @marcenacp is there a better way to link PRs to issues? :)

python/ml_croissant/ml_croissant/_src/datasets_test.py
```diff
@@ -10,84 +10,31 @@
...
 def get_error_msg(folder):
     with open(f"{folder}/output.txt", "r") as file:
         return file.read().strip()
 
 
-# TODO(https://github.com/mlcommons/croissant/issues/14): Progressively move tests from
-# test_static_analysis_old to test_static_analysis
-@pytest.mark.parametrize("folder", ["distribution_bad_contained_in"])
+@pytest.mark.parametrize("folder", [
+                          # Distribution.
```
marcenacp, line 19: Should you locate every folder with a regular expression, so that adding a new folder is transparent?

marcenacp (approved):
Thank you, Stanze! The input/output is much more readable in files like this.

ccl-core:
Fixes #14

python/ml_croissant/ml_croissant/_src/datasets_test.py
ccl-core: Hi! Not sure what you mean with locating each folder w. a regex?
Another option might be to just iterate over tests/graphs... but I like how we are writing down all tests in the current version, grouping similar tests together and adding comments, it makes it easier to read I think?

PR #176, opened 2023-08-22
Use constants instead of strings when expanding JSON-LDs.

marcenacp:


ccl-core (approved):
The codebase looks much cleaner now, thanks!

PR #188, opened 2023-08-30
Add coco2014 contentSize

mkuchnik:
Some fields, including contentSize, have filler values (e.g., logically null). This PR removes null values for coco2014 contentSize.

### DIFF
--- a/datasets/coco2014/metadata.json
+++ b/datasets/coco2014/metadata.json
@@ -46,7 +46,7 @@
     {
       "@type": "sc:FileObject",
       "name": "train2014.zip",
-      "contentSize": " B",
+      "contentSize": "13510573713 B",
       "contentUrl": "http://images.cocodataset.org/zips/train2014.zip",
       "encodingFormat": "application/zip",
       "sha256": "sha256"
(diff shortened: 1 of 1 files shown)


mkuchnik:
recheck

marcenacp (approved):
Thanks!

Eventually, should it be an integer always in bytes (machine readable) or a string (human readable)?

marcenacp:
> recheck

When I push, I noticed the cla-check gets re-run.

mkuchnik:
recheck

mkuchnik:
> Thanks!
> 
> Eventually, should it be an integer always in bytes (machine readable) or a string (human readable)?

I think machine readable makes sense for exporting to a human-readable front-end. It also avoids confusion from different units (MB vs. MiB).

> > recheck
> 
> When I push, I noticed the cla-check gets re-run.

I don't think I can change anything from my end with regards to merging. The cla-bot displays the same log from before the `recheck`.

benjelloun:
> I think machine readable makes sense for exporting to a human-readable front-end. The length in bytes can serve as a form of file validation in place of a hash; however, human-readable forms will lose some precision through rounding errors. Machine readable also avoids confusion from different units (MB vs. MiB).

Unfortunately, schema.org/contentSize is a bit under-specified, and we observe a wide variety of values used in the wild. In the Croissant spec, we can provide a more precise recommendation on how contentSize values should be formatted.

pierrot0 (approved):

mkuchnik:
recheck

PR #190, opened 2023-08-30
Remove overloaded reference to label

mkuchnik:
The spec is using `label` to mean a machine learning output variable. This PR is to remove references to `label` in a more general context.


marcenacp (approved):
Thanks!

PR #193, opened 2023-08-31
Make `--record_set` mandatory when loading and check that it exists.

marcenacp:


python/ml_croissant/ml_croissant/_src/core/git_test.py
```diff
@@ -11,29 +11,35 @@
...
 from ml_croissant._src.core.path import Path
 
-_GIT_LFS_CONTENT = """version https://git-lfs.github.com/spec/v1
+_GIT_LFS_CONTENT = lambda: """version https://git-lfs.github.com/spec/v1
 oid sha256:5e2785fcd9098567a49d6e62e328923d955b307b6dcd0492f6234e96e670772a
 size 309207547
 """
 
-_NON_GIT_LFS_CONTENT = """name,age
+_NON_GIT_LFS_CONTENT = lambda: """name,age
```
ccl-core, line 19: Just asking, isn't it _GIT_NON_LFS ?

python/ml_croissant/ml_croissant/_src/datasets.py
```diff
@@ -65,6 +65,8 @@ def __post_init__(self):
 
     def records(self, record_set: str) -> Records:
         """Accesses all records belonging to the RecordSet named `record_set`."""
```
ccl-core, line 67: Maybe add "if it exists" to the function comment?

ccl-core (approved):
Thank you Pierre!

python/ml_croissant/ml_croissant/_src/datasets.py
marcenacp: Done.

python/ml_croissant/ml_croissant/_src/core/git_test.py
marcenacp: git-lfs goes in one word (https://git-lfs.com). I wanted to convey this file is not a git-lfs file, because it doesn't have the proper header, so I called it non git-lfs.

Does it make sense? If not, we can always change later for a better wording.

marcenacp:
Thanks, @ccl-core!

PR #195, opened 2023-08-31
Make `git` an optional dependency of Croissant

marcenacp:
Fixes: https://github.com/mlcommons/croissant/issues/192

### DIFF
--- a/python/ml_croissant/ml_croissant/_src/core/optional.py
+++ b/python/ml_croissant/ml_croissant/_src/core/optional.py
@@ -60,6 +60,24 @@ class OptionalDependencies(object):
     lazily imported here.
 
     Warning: we haven't found a way to properly type the return of each property.
+
+    Usage:
+
+    - Add a new optional dependency below:
+
+    ```python
+    @cached_class_property
+    def foo(cls):
+        return _try_import("goo")
+    ```
+
+    - Use this new dependency:
+
+    ```python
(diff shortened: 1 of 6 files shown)


python/ml_croissant/ml_croissant/_src/core/optional.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class OptionalDependencies(object):
+    """Optional dependencies can be heavy and need not be loaded by all clients.
+
+    Some datasets require optional dependencies for data generation. To allow for
+    the default installation to remain lean, those heavy dependencies are
+    lazily imported here.
+
+    Warning: we haven't found a way to properly type the return of each property.
```
ccl-core, line 62: This might be a TODO(issue) ?

ccl-core (approved):
That's great, thank you Pierre!

I think it'd be good here to also have a comment somewhere for future developers stating how to add an optional dependency?

python/ml_croissant/ml_croissant/_src/core/optional.py
marcenacp: > I think it'd be good here to also have a comment somewhere for future developers stating how to add an optional dependency?

Good idea, thanks! Done.

> This might be a TODO(issue) ?

I don't think it's an issue worth spending time on. For instance, tfds has the same problem. I only wanted to underline it here.

PR #198, opened 2023-09-06
Read images in MNIST as PIL.Image.Image

marcenacp:
After this PR, I have to continue with the issue https://github.com/mlcommons/croissant/issues/199, in order to have it working for all images.


python/ml_croissant/scripts/load.py
```diff
@@ -57,7 +56,27 @@ def main(argv):
     num_records = FLAGS.num_records
     debug = FLAGS.debug
     update_output = FLAGS.update_output
+    return load(
```
ccl-core, line 59: I wonder if this change would have been better placed in a separated PR, given that is not related to Image processing?

ccl-core (approved):
Very cool, thank you!

PR #203, opened 2023-09-06
`ml:transform` accepts "jsonPath"

marcenacp:
Fixes: https://github.com/mlcommons/croissant/issues/199

This allows to read JSON within any table, e.g. in the column of a Parquet file as it is the case for images in Hugging Face.


python/ml_croissant/ml_croissant/_src/structure_graph/nodes/source.py
```diff
@@ -103,11 +106,13 @@ def from_jsonld(cls, issues: Issues, jsonld: Json) -> list[Transform]:
                 )
                 continue
             format = transform.get(constants.ML_COMMONS_FORMAT)
+            json_path = transform.get(constants.ML_COMMONS_JSON_PATH)
             regex = transform.get(constants.ML_COMMONS_REGEX)
             replace = transform.get(constants.ML_COMMONS_REPLACE)
             separator = transform.get(constants.ML_COMMONS_SEPARATOR)
             if (
                 format is None
```
ccl-core, line 114: if we expect this to grow we could also `if all(v is None for v in [format, json_path, ...])``

ccl-core (approved):
Thanks!

python/ml_croissant/ml_croissant/_src/structure_graph/nodes/source.py
marcenacp: Done.

PR #204, opened 2023-09-06
Remove tracing back the package name in mlcroissant

marcenacp:
1. This removes code...
2. This won't work once the package is built and downloaded eg from PyPI...

So better simplify now.


ccl-core (approved):
Thanks!

PR #205, opened 2023-09-06
Migrate `csvColumn`->`column` for use with all columnar formats.

marcenacp:
Fixes: https://github.com/mlcommons/croissant/issues/200.


ccl-core (approved):
Thanks!

PR #207, opened 2023-09-07
Change GroupRecordSet call to return the sample directly.

ccl-core:
Croissant now output samples in the form: `{"name": "person0", "age": 0}`} (instead of: `{"persons": {"name": "person0", "age": 0}}`, as it was before)


ccl-core:
Hi Pierre, I'm still having a look at why the data generation pipeline seems to be broken in my codespaces. But the CI Unittests seem to be green.

python/ml_croissant/ml_croissant/_src/operation_graph/operations/group.py
```diff
@@ -14,4 +14,4 @@ class GroupRecordSet(Operation):
     def __call__(self, *fields: pd.Series):
         """See class' docstring."""
         concatenated_series = {k: v for field in fields for k, v in field.items()}
```
marcenacp, line 16: nit: `return {...}`

marcenacp (approved):
Thanks!

PR #209, opened 2023-09-07
Rename `ml_croissant`->`mlcroissant` before publishing to PyPI.

marcenacp:


ccl-core (approved):
Thanks, mlcroissant is def nicer!

PR #222, opened 2023-09-20
Monjishb

monke6942021:
Add the requirements.txt file, and maybe the peoples-speech metadata. I will have to improve the peoples-speech metadata so that it can be used for accessing records.


requirements.txt
```diff
@@ -0,0 +1,29 @@
+absl-py==1.4.0
```
marcenacp, line 1: mlcroissant uses [pyproject.toml](https://github.com/mlcommons/croissant/blob/main/python/mlcroissant/pyproject.toml) rather than requirements.txt.

mini_CC_BY_SA/metadata.json
```diff
@@ -0,0 +1,91 @@
+{
```
marcenacp, line 1: All datasets should go to [`croissant/datasets/`](https://github.com/mlcommons/croissant/tree/main/datasets).

marcenacp: You can put external data in a separate `data` folder. Example: https://github.com/mlcommons/croissant/tree/main/datasets/coco2014-mini.

marcenacp (changes requested):
Thanks!

PR #226, opened 2023-09-25
Separating hermetic and nonhermetic tests

ccl-core:
Certain tests are non-hermetic as they need to download data from the Internet. We separate them from hermetic tests.

### DIFF
--- a/python/mlcroissant/pyproject.toml
+++ b/python/mlcroissant/pyproject.toml
@@ -68,7 +68,7 @@ requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
 
 [tool.pytest.ini_options]
-addopts = --strict-markers
+addopts = "--strict-markers"
 markers = [
     "nonhermetic: Non-hermetic test cases (data from the internet)."
 ]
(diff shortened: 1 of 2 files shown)


python/mlcroissant/pyproject.toml
```diff
@@ -66,3 +66,8 @@ exclude = ["recipes*"]
 [build-system]
 requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
+
+[tool.pytest.ini_options]
+markers = [
```
marcenacp, line 71: According to the documentation (https://docs.pytest.org/en/latest/how-to/mark.html), you can also add: `addopts = --strict-markers`

python/mlcroissant/pyproject.toml
```diff
@@ -66,3 +66,8 @@ exclude = ["recipes*"]
...
+    "nonhermetic: Non-hermetic test cases (data from the internet)."
```
marcenacp, line 72: Note: to launch hermetic tests, we have to launch `pytest -m "not nonhermetic"`. I don't like the double negation, but I don't see a better name. WDYT?

marcenacp (approved):
Thanks!

python/mlcroissant/pyproject.toml
ccl-core: Done.

PR #236, opened 2023-10-10
Support both local and distant files.

marcenacp:
I also added support for more file types.

End-to-end tests should come soon.


ccl-core (approved):
Thanks!

PR #241, opened 2023-10-13
Allow to specify a FileObject.containedIn with a contentUrl of a relative path.

marcenacp:
Fixes: https://github.com/mlcommons/croissant/issues/234

In this PR, we also simplify Download(node=node, url=url)->Download(node=node).


ccl-core (approved):
Nice!

PR #250, opened 2023-10-19
Add POC config visualizer

aidazolic:
- Adds a POC visualizer for the generated config, currently only showing the metadata.
- Modifies the wizard UI to use tabs instead of buttons.


aidazolic:
recheck

marcenacp (commented):
Thanks a lot for the PR! In the meantime, we iterated on the UX, I'll keep you updated when you come back from leave.

PR #254, opened 2023-10-19
Update git_test.py

ccl-core:
To solve `ImportError: Failed to initialize: Bad git executable.` error

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/core/git_test.py
+++ b/python/mlcroissant/mlcroissant/_src/core/git_test.py
@@ -5,10 +5,10 @@ import tempfile
 from unittest import mock
 
 from etils import epath
-import git
 
 from mlcroissant._src.core.git import download_git_lfs_file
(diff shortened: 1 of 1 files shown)


python/mlcroissant/mlcroissant/_src/core/git_test.py
```diff
@@ -46,8 +46,10 @@ def test_download_git_lfs_file():
         filepath=epath.Path("/tmp/full/path.json"),
         fullpath=pathlib.PurePath("path.json"),
     )
+    git = deps.git
     with mock.patch.object(git, "Git", autospec=True) as git_mock:
         download_git_lfs_file(file)
+        git = deps.git
```
marcenacp, line 52: Remove?

ccl-core: Thanks for spotting, done.

PR #258, opened 2023-10-19
Stricter pytype rules in pyproject.toml.

marcenacp:
With these parameters, errors that used to not fail now fail (e.g., the errors fixed in https://github.com/mlcommons/croissant/pull/248).


ccl-core (approved):
Thank you a lot Pierre!

PR #260, opened 2023-10-20
Only execute needed operations for performance reasons.

marcenacp:
If we have the following graph:

```
                       InitOperation
      _______________________|___________________
      |                      |                   |
 Download(file1)      Download(file2)      Download(file3)
   Read(file1)          Read(file2)          Read(file3)
GroupRecordSet(1)    GroupRecordSet(2)    GroupRecordSet(3)
      |                      |                   |
     ...                    ...                 ...
```

and you're interested only in `GroupRecordSet(2)`, you only want to execute operations in its ancestors.


ccl-core (approved):
Makes sense, thanks Pierre!

PR #261, opened 2023-10-20
Update postcreatecommand

ccl-core:
This is needed after: https://github.com/mlcommons/croissant/commit/ef3955808f6ebf506246943886062538caf7d0e6

### DIFF
--- a/.devcontainer/postCreateCommand.sh
+++ b/.devcontainer/postCreateCommand.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+sudo apt-get update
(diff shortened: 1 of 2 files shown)


.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+sudo apt-get update
+sudo apt-get install -y libgraphviz-dev
+pip3 install python/mlcroissant/.[dev]
```
marcenacp, line 5: nit: carriage return.

marcenacp (approved):
Thanks!

PR #262, opened 2023-10-20
Fix issue when several FileObjects are extracted from a FileObject.

marcenacp:


ccl-core (approved):
Thanks!

datasets/recipes/file_object_in_zip.json
```diff
@@ -44,6 +44,14 @@
       "encodingFormat": "application/zip",
       "sha256": "85c13a7e6940a8376f2448ed6078d4fa407ae33c7839e6d041a0e34413366792"
     },
+    {
+      "@type": "sc:FileObject",
+      "name": "other_csv_in_zip",
+      "description": "Do not remove! This is used in hermetic integration tests. There used to be a problem where several FileObjects could not be read from the same FileObject.",
+      "containedIn": "zip_with_csv",
+      "contentUrl": "csv_in_zip.csv",
```
goeffthomas, line 52: @marcenacp Was this meant to be the same `contentUrl`? Seems a little weird to have 2 `FileObject` representations of the same file in `distribution`.

marcenacp: @goeffthomas, the goal was to emulate as if there were two files... But I agree we can do better and actually have several zipped files. I have a pending PR.

PR #271, opened 2023-10-26
Always manipulate bytes rather than str and unit test _cast_value.

marcenacp:


python/mlcroissant/mlcroissant/_src/tests/records.py
```diff
@@ -11,15 +10,16 @@
 def record_to_python(record: Any):
```
mkuchnik, line 10: Underscore name for internal API (e.g., `_record_to_python`)?

mkuchnik (approved):

python/mlcroissant/mlcroissant/_src/tests/records.py
marcenacp: Thanks for the review!

By default, all functions/classes are internal because they are in a folder called `_src`. So a user trying to import directly a function would (hopefully) notice they are doing something wrong: `from mlcroissant._src import my_function`.

The main [`__init__`](https://github.com/mlcommons/croissant/blob/main/python/mlcroissant/mlcroissant/__init__.py) takes care of exposing the external API.

Until now, only internal functions to a single file starts with an underscore, i.e. when a function is not shared between several modules.

Of course, this design is open to discussion. WDYT?

marcenacp: @mkuchnik
(Michael, I will merge now as several PRs start depending on this PR. But I am happy to continue the discussion offline and push patches if needed.)

PR #272, opened 2023-10-26
First step in removing string usage in mlc.

marcenacp:
This PR exposes `DataTypes` and `EncodingFormat` in mlcroissant.

It is only a first step towards having better typing/DX experience. In future PRs, we could use [enums](https://docs.python.org/3/library/enum.html) and propagate the correct types in the codebase ([example](https://github.com/mlcommons/croissant/blob/d04dc82820f0a3295f759e8c14ae36e8a74dd4c7/python/mlcroissant/mlcroissant/_src/structure_graph/nodes/file_object.py#L28)).

At least developers can now write:

```python
encoding_format=mlc.EncodingFormat.GIT
```


ccl-core (approved):
Nice! Thank you Pierre

PR #273, opened 2023-10-26
Fix make tests on Mac

mkuchnik:
`make tests` currently assumes `/proc/cpuinfo` exists. On Macs, this is not the case, and `sysctl` can be used instead.

This switch per OS should probably be removed eventually, or implemented in a more robust manner.


ccl-core (approved):
Thanks!

PR #277, opened 2023-10-27
Fix test loop with pytest parametrize

mkuchnik:
Per https://github.com/mlcommons/croissant/pull/275#discussion_r1374716066


marcenacp (approved):
Thanks!

PR #278, opened 2023-10-30
Read files line by line with {"fileProperty": "lines"}.

marcenacp:


python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
```diff
@@ -60,6 +60,21 @@ def _read_file_content(self, encoding_format: str, file: Path) -> pd.DataFrame:
                         " mlcroissant[parquet]`."
                     ) from e
             elif encoding_format == "text/plain":
+                fields = self.fields
+                # Note: This is an approximation. It could be that the same FileObject
```
ccl-core, line 64: Shall we link this to an issue for future tracking?

ccl-core (approved):
Thanks Pierre!

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
marcenacp: Good idea! I think I can even push an explicit error message, saying what to do in this case. Otherwise, I'm afraid the issue will get lost in the backlog.

PR #279, opened 2023-10-30
Separate MIME type from reading method and add a clear error message.

marcenacp:
This is a follow-up PR after Costanza's review: https://github.com/mlcommons/croissant/pull/278#discussion_r1376171734

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
+++ b/python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
@@ -1,6 +1,7 @@
 """Read operation module."""
 
 import dataclasses
+import enum
 import json
 
 from etils import epath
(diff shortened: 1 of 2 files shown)


python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
```diff
@@ -19,6 +20,50 @@
...
+    if len(reading_methods) == 0:
+        return ReadingMethod.NONE
+    if len(reading_methods) > 1:
+        raise ValueError(
+            f"Cannot read {node=}. The fields use several reading methods:"
+            f" {reading_methods}. Reading the same FileObject/FileSet using different"
+            " reading methods has yet to be implemented. Please, create an issue"
+            " (https://github.com/mlcommons/croissant/issues/new) if your dataset"
+            " requires this feature. Alternatively, you can use two different"
+            " FileObject/FileSet pointing to the same resource."
```
ccl-core, line 62: Do we already have any croissant dataset which does so -- to cite as an example?

ccl-core (approved):
Very nice, thanks Pierre!

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
marcenacp: No, we don't. My assumption is that there won't be such a dataset, and worst case we can still create two different FileObjects/FileSets.

PR #282, opened 2023-10-31
Support reading binary data from a zip.

marcenacp:
Fixes a few issues to read binary files contained in a zip:

- When ["streaming" a dataset](https://github.com/mlcommons/croissant/blob/255df419fb05efc96e4cf7ff23f93e76b5948184/python/mlcroissant/mlcroissant/_src/operation_graph/execute.py#L76), we used to not have the same graph as in https://github.com/mlcommons/croissant/pull/260.
- Now, ReadField outputs a pd.Series (rather than a scalar).
- Refactored some code.


marcenacp:
> Thanks! Before merging, could you maybe update the PR description to include more context? In case we need it for future debugging

Thanks, it's done :)

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field_test.py
```diff
@@ -53,3 +54,13 @@ def test_cast_value_image(open_mock):
     expected = field._cast_value(b"PNG...Some image...", DataType.IMAGE_OBJECT)
     open_mock.assert_called_once()
     assert expected == "opened_image"
+
+
+def test_extract_lines():
+    with tempfile.TemporaryDirectory() as tempdir:
+        path = tempdir + "/file.txt"
+        with open(path, "wb") as f:
+            f.write(b"bonjour\nh\xc3\xa9llo\n")
```
mkuchnik, line 63: Is Croissant's "lines" behavior supposed to be the same as Python's [splitlines](https://docs.python.org/3/library/stdtypes.html#str.splitlines)? If yes, should tests cover '\r' and other characters?

marcenacp: Done.

I thought `splitlines` was only for str, but it plays nicely with bytes as well. I also updated the unit test (even though now it almost boils down to testing `splitlines` itself). Thanks!

PR #287, opened 2023-11-01
Add the `lineNumbers` if at least one field requires it.

marcenacp:
Previously, only "text/plain" could be enriched with the `lineNumbers`.


mkuchnik (approved):
Works on my end. Thanks!

PR #288, opened 2023-11-02
Add FLORES 200 test set

mkuchnik:
FLORES 200 dataset contained training set only. Now it has the test set.


datasets/flores-200/metadata.json
```diff
@@ -215,6 +231,152 @@
           }
         }
       ]
+    },
+    {
+      "@type": "ml:RecordSet",
```
marcenacp, line 236: Note: An alternate approach would have been to specify `"split": "dev" | "devtest"` as a parameter of all ml:RecordSets in order not to repeat them.

But because filtering is not implemented yet in mlcroissant, your current approach is probably the best.

marcenacp (approved):
Thanks1

PR #291, opened 2023-11-03
Only create a `Join` if it is needed.

marcenacp:
We used to always create a `Join` operation even if no join is needed.


ccl-core (approved):
Thanks!

PR #298, opened 2023-11-03
Add git lfs install instructions

mkuchnik:
There seems to be implicit usage of `git lfs`, which is potentially surprising and can cause flaky tests for unsuspecting users. The patch adds a check and install instructions.

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/core/git.py
+++ b/python/mlcroissant/mlcroissant/_src/core/git.py
@@ -34,4 +34,9 @@ def download_git_lfs_file(file: Path):
     logging.info(
         "Downloading git-lfs file: %s in working dir: %s", fullpath, working_dir
     )
-    repo.execute(["git", "lfs", "pull", "--include", fullpath])
+    try:
+        repo.execute(["git", "lfs", "pull", "--include", fullpath])
+    except deps.git.exc.GitCommandError as ex:
+        if ("You called a Git command named 'lfs', which does not exist."
+                in ex.stderr):
+            raise RuntimeError("git lfs is not installed!") from ex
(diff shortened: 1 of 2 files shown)


python/mlcroissant/mlcroissant/_src/core/git.py
```diff
@@ -34,4 +34,9 @@ def download_git_lfs_file(file: Path):
     logging.info(
         "Downloading git-lfs file: %s in working dir: %s", fullpath, working_dir
     )
-    repo.execute(["git", "lfs", "pull", "--include", fullpath])
+    try:
+        repo.execute(["git", "lfs", "pull", "--include", fullpath])
+    except deps.git.exc.GitCommandError as ex:
+        if ("You called a Git command named 'lfs', which does not exist."
+                in ex.stderr):
+            raise RuntimeError("git lfs is not installed!") from ex
```
marcenacp, line 42: This implementation only relies on the string of the error message in the GitPython library. I would rather always re-raise the exception with an explicit error message.

```python
    try:
        repo.execute(["git", "lfs", "pull", "--include", fullpath])
    except deps.git.exc.GitCommandError as e:
        raise RuntimeError(f"Problem when launching `git lfs`. Possible problems: Have you installed git lfs locally? Is {fullpath} a valid `git lfs` repository?") from e
```

WDYT?

marcenacp (changes requested):

python/mlcroissant/mlcroissant/_src/core/git.py
mkuchnik: Looks good to me! Updated.

marcenacp (approved):
Thanks!

PR #309, opened 2023-11-07
[Issue # 295] Replace GroupRecordSetStart/GroupRecordSetEnd by ReadFields.

marcenacp:
This allows to have all fields computed at the same place. For example, this avoids repeating computation to compute separately lines and lineNumbers. Now, all fields are in the same `ReadFields.__call__` function.

Fixes: https://github.com/mlcommons/croissant/issues/295

Note for reviewers: the main commit is the first commit of the chain. Then we remove code, rename, comment and test.


ccl-core (approved):
Thanks Pierre!

PR #312, opened 2023-11-07
checking sha256 and md5 at download

St3V0Bay:
Hi there,

- sha256 and md5 checking is included
- tested on my end and it works
- PyTest **not written** yet

Edge cases worth discussing: 
(1) if sha256 and md5 hash are provided via json, which one to take (or both)?
(2) Should checksums be mandatory (i.e. do not start download if not provided)?

Please review code and advise.

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/operation_graph/operations/download.py
+++ b/python/mlcroissant/mlcroissant/_src/operation_graph/operations/download.py
@@ -117,6 +117,20 @@ def extract_git_info(full_url: str) -> tuple[str, str | None]:
             f" {_SUPPORTED_HOSTS}. Contact the Croissant team to support more hosts."
         )
 
+def _get_hash_type(node):
+    result=None
+    if (node.md5 is not None) and (node.sha256 is not None):
+        logging.info("self.node has md5 and sha256 key")
+        result =  "md5 and sha256"
+    elif node.md5 is not None:
+        logging.info("self.node has md5 key")
+        result = "md5"
+    elif node.sha256 is not None:
+        logging.info("self.node has sha256 key")
+        result = "sha256"
+    else:
+        pass
+    return result
 
 @dataclasses.dataclass(frozen=True, repr=False)
 class Download(Operation):
@@ -128,6 +142,17 @@ class Download(Operation):
     def _download_from_http(self, filepath: epath.Path):
         response = requests.get(self.url, stream=True, timeout=10)
         total = int(response.headers.get("Content-Length", 0))
+
+        hash_type = _get_hash_type(self.node)
+
+        if hash_type == "sha256":
+            hash_eng = hashlib.sha256()
+        elif hash_type == "md5":
+            hash_eng = hashlib.md5()
+        else:
+            pass
+            #TODO: What to do if both hash values are given
+
         with filepath.open("wb") as file, tqdm.tqdm(
             desc=f"Downloading {self.url}...",
             total=total,
@@ -137,8 +162,16 @@ class Download(Operation):
         ) as bar:
             for data in response.iter_content(chunk_size=_DOWNLOAD_CHUNK_SIZE):
                 size = file.write(data)
+                hash_eng.update(data)
                 bar.update(size)
 
+        downloaded_file_hash = hash_eng.hexdigest()
+
+        if downloaded_file_hash != getattr(self.node,hash_type):
+            logging.info("Hash of downloaded file is not identical with reference in metadata.json")
+            os.remove(filepath)
+            raise ValueError('Hash of downloaded file is not identical with reference in metadata.json')
+            
     def _download_from_git(self, filepath: epath.Path):
         username = os.environ.get(constants.CROISSANT_GIT_USERNAME)
         password = os.environ.get(constants.CROISSANT_GIT_PASSWORD)


python/mlcroissant/mlcroissant/_src/operation_graph/operations/download.py
```diff
@@ -128,6 +142,17 @@ class Download(Operation):
     def _download_from_http(self, filepath: epath.Path):
         response = requests.get(self.url, stream=True, timeout=10)
         total = int(response.headers.get("Content-Length", 0))
+
+        hash_type = _get_hash_type(self.node)
```
marcenacp, line 146: This would then be simplified to `hash = _get_hash(self.node)`

Then you can use `hash.update(data)`

python/mlcroissant/mlcroissant/_src/operation_graph/operations/download.py
```diff
@@ -137,8 +162,16 @@ def _download_from_http(self, filepath: epath.Path):
         ) as bar:
             for data in response.iter_content(chunk_size=_DOWNLOAD_CHUNK_SIZE):
                 size = file.write(data)
+                hash_eng.update(data)
                 bar.update(size)
 
+        downloaded_file_hash = hash_eng.hexdigest()
+
+        if downloaded_file_hash != getattr(self.node,hash_type):
```
marcenacp, line 170: This if condition will probably fail in the unit tests. Can you fix the tests?

python/mlcroissant/mlcroissant/_src/operation_graph/operations/download.py
```diff
@@ -128,6 +142,17 @@ class Download(Operation):
...
+
+        hash_type = _get_hash_type(self.node)
+
+        if hash_type == "sha256":
+            hash_eng = hashlib.sha256()
+        elif hash_type == "md5":
+            hash_eng = hashlib.md5()
+        else:
+            pass
+            #TODO: What to do if both hash values are given
```
marcenacp, line 154: This shouldn't happen ([source](https://github.com/mlcommons/croissant/blob/ec5dc76e4c20d28acf681c926799ee763b855146/python/mlcroissant/mlcroissant/_src/structure_graph/nodes/file_object.py#L39)), but I still added a `ValueError` in the function above.

marcenacp (changes requested):
Thanks! If you fix the tests, we should be able to merge soon. :)

python/mlcroissant/mlcroissant/_src/operation_graph/operations/download.py
```diff
@@ -117,6 +117,20 @@ def extract_git_info(full_url: str) -> tuple[str, str | None]:
             f" {_SUPPORTED_HOSTS}. Contact the Croissant team to support more hosts."
         )
 
+def _get_hash_type(node):
```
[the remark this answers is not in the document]
St3V0Bay: fully agree. looks much cleaner.

python/mlcroissant/mlcroissant/_src/operation_graph/operations/download.py
St3V0Bay: I ran a few tests. And it asserts that "at least one hash" is provided, which needs to be changed to "exactly one hash". 

If 2 hashes are provided, the script continues with whatever the following function returns (due to order of if statement it is "md5" currently.

```
def _get_hash(file_object: FileObject):
    if file_object.md5:
        return hashlib.md5()
    elif file_object.sha256:
        return hashlib.sha256()
    else:
        raise ValueError("md5 and sha256 are not specified.")
```

St3V0Bay:
recheck

python/mlcroissant/mlcroissant/_src/operation_graph/operations/download.py
```diff
@@ -117,6 +117,13 @@ def extract_git_info(full_url: str) -> tuple[str, str | None]:
             f" {_SUPPORTED_HOSTS}. Contact the Croissant team to support more hosts."
         )
 
+def _get_hash_obj(file_object: FileObject):
```
marcenacp, line 120: `_get_hash` to avoid abbreviations?

python/mlcroissant/mlcroissant/_src/operation_graph/operations/download_test.py
```diff
@@ -68,3 +75,38 @@ def test_insert_credentials():
...
+
+def test_get_hash_obj_sha256():
+    file_object = FileObject(
+            sha256="12345",
+        )
+    hash_obj = _get_hash_obj(file_object)
+
+    assert isinstance(hash_obj, type(hashlib.sha256()))
+
+def test_checking_hash_correctly():
```
marcenacp, line 95: I see two possible issues with this unit test:

- The test is non-hermetic (meaning it downloads data from the internet).
- The test checks the function by using the content of the function itself (you copy/pasted the logic from the code to the test).

Typically we would use mocking instead ([see the documentation](https://docs.python.org/3/library/unittest.mock.html)).

Warning: this is "pseudo-code" which I haven't tested.

```python
@mock.patch.object(requests, 'get')
def test_checking_hash_correctly(get_hash_mock):
  # Assert that the file doesn't exist
  # assert ...
  node = FileObject(
    sha256='c617db2c7470716250f6f001be51304c76bcc8815527ab8bae734bdca0735737',
  )
  Download(node=node)()
  # Assert that the file has been written in the good path with the proper hash 
  # assert ...
```

Also, other tests in the codebase should fail. So I recommend you focus on those tests. We can write a unit test later in another PR if needed.

marcenacp (changes requested):

python/mlcroissant/mlcroissant/_src/operation_graph/operations/download.py
St3V0Bay: There is a function called get_hash() (written by you) to get URL hashed. I thought that would be too similar to _get_hash(). May be _get_hash_algorithm() or _get_hash_engine() (not sure whether it's an "engine"). What do you think?

marcenacp: Thanks, `_get_hash_algorithm ` LGTM. 🙂

marcenacp:
Note: Before GitHub Actions let you execute the tests, you can launch them locally (`make pytest`).

St3V0Bay:
@marcenacp - i think we are good here. I'd say we can merge

marcenacp:
@St3V0Bay Thanks!

PR #313, opened 2023-11-07
Support `ml:BoundingBox` in Croissant with the example of COCO2014.

marcenacp:


python/mlcroissant/mlcroissant/_src/core/ml/bounding_box.py
```diff
@@ -0,0 +1,53 @@
...
+import enum
+from typing import Any
+
+
+# TODO(marcenacp): Coco supports XYWH.
+class BoundingBoxFormat(enum.Enum):
+    """The format of the bounding box to extract from.
+
+    Note:
+    - Croissant uses XYXY as the common format.
```
ccl-core, line 12: TODO?

ccl-core (approved):
Thanks Pierre!

python/mlcroissant/mlcroissant/_src/core/ml/bounding_box.py
marcenacp: I chose to follow [Keras' bounding box formats](https://keras.io/api/keras_cv/bounding_box/formats). Croissant will have a standard output format (not sure yet whether it will be `XYXY` or `XYWH`).

marcenacp: I ended up removing entirely this class: `mlcroissant` won't check for the format for now.

PR #314, opened 2023-11-07
Cross obj data loading

### LINKED ISSUE #281: Upload a JSON-LD file and edit it (1 pt)
See comments on https://github.com/mlcommons/croissant/pull/280.

~~We use `mlc.Metadata` as the source of truth.~~

Edit:

We decided to use our own simple abstraction `Metadata`/`FileSet`/etc as `mlc.Metadata` panics if the metadata are not complete (which is the expected behaviour). These abstractions are currently defined in https://github.com/mlcommons/croissant/blob/255df419fb05efc96e4cf7ff23f93e76b5948184/wizard/st_state.py#L26 (only `Metadata` is fully defined to this date).

drayuka:
Closes #281


wizard/state.py
```diff
@@ -31,17 +28,29 @@ class CurrentStep:
...
 
 
-class Files:
-    pass
-
-
-class RecordSets:
-    pass
+@dataclasses.dataclass
+class Distribution:
```
marcenacp, line 32: All classes should have docstrings.

wizard/state.py
```diff
@@ -53,3 +62,75 @@ class Metadata:
...
+    def remove_distribution(self, key: int) -> None:
+        del self.distributions[key]
+    def add_record_set(self, record_set: RecordSet) -> None:
+        self.record_sets.append(record_set)
+    def update_record_set(self, key: int, record_set: RecordSet) -> None:
+        self.record_sets[key] = record_set
+    def remove_record_set(self, key: int) -> None:
+        del self.record_sets[key]
+
+def CanonicalToWizard(dataset: mlc.Dataset) -> Croissant:
```
marcenacp, line 86: This could be a [`classmethod`](https://docs.python.org/3/library/functions.html#classmethod) of `Croissant` and be called `from_canonical`.

Like this you'd write:

```python
croissant = Croissant.from_canonical(dataset)
```

wizard/state.py
```diff
@@ -53,3 +62,75 @@ class Metadata:
 
     def __bool__(self):
         return self.name != "" and self.url != ""
+    
+@dataclasses.dataclass
+class Croissant:
+    metadata: Metadata = Metadata()
+    distributions: list[Distribution] = dataclasses.field(default_factory=list)
```
marcenacp, line 69: `distributions`/`record_sets` are under Metadata in mlcroissant. Should that also be the case here in the editor? I.e. should we follow the same structure to limit the overhead of learning two representations?

Note: in Croissant, distribution is singular because of https://schema.org/distribution. Files or resources could be alternative names that do not alter the original meaning of `distribution`.  Like this you can also write later:

```python
for resource in croissant.resources:

# or

for file in croissant.files:

# or

for file in croissant.distribution:
```

wizard/state.py
```diff
@@ -53,3 +62,75 @@ class Metadata:
...
+        name=canonical_metadata.name,
+        description=canonical_metadata.description,
+        citation=canonical_metadata.citation,
+        license=canonical_metadata.license,
+        url=canonical_metadata.url
+    )
+    distributions = []
+    for file in canonical_metadata.distribution:
+        if isinstance(file, mlc.nodes.FileObject):
+            distributions.append(Distribution(
```
marcenacp, line 98: Here we see that it's interesting to have a distinction between FileObject and FileSets. If you keep the names `FileObject`/`FileSet` and copy mlcroissant, you reduce the mental overload and you allow easy checks:

```
if isinstance(file, FileObject):

# or

if isinstance(file, FileSet):
```

Probably FileObject/FileSet will be handled by different Streamlit components so this distinction would be interesting.

marcenacp (changes requested):
Thanks for the PR, LGTM!

I added a few comments. Mostly, I feel the Streamlit state should mirror mlcroissant state.

```
Metadata -> Metadata
Metadata.distribution -> Metadata.distribution
FileObject -> FileObject
etc
```

Arguments:

- This removes the overhead of learning a new representation.
- We use the same vocabulary to refer to things.
- It becomes a 1-to-1 mapping.

WDYT?

wizard/state.py
drayuka: At some point I'd like to talk to the people designing this stuff, as they and I have some disagreements on how structured data should be shaped, done though.

wizard/state.py
drayuka: added both, however, for the moment, it seems like everything addable from the application is a FileObject, but this was the conversation we were having this morning I believe.

python/mlcroissant/mlcroissant/__init__.py
```diff
@@ -17,4 +18,5 @@
     "nodes",
     "Records",
     "ValidationError",
+    "FileObject",
```
marcenacp, line 21: nit: Alphabetical order.

wizard/app.py
```diff
@@ -1,12 +1,24 @@
...
+def init_state():
+
+    if Metadata not in st.session_state:
+        st.session_state[Metadata] = Metadata()
+
+    if mlc.Dataset not in st.session_state:
+        st.session_state[mlc.Dataset] = None
+
+    if CurrentStep not in st.session_state:
+        st.session_state[CurrentStep] = "start"
```
marcenacp, line 20: Better not to use strings when manipulating the state. WDYT?

wizard/core/files.py
```diff
@@ -7,18 +7,7 @@
...
-    name: str = ""
-    description: str | None = None
-    content_url: str = ""
-    encoding_format: str | None = None
-    sha256: str | None = None
-    df: pd.DataFrame | None = None
-
-    def __post_init__(self):
-        self.df = self.df.infer_objects()
+from .state import FileObject
```
marcenacp, line 10: In the whole repo, we use absolute imports relative to the root (i.e., where the pyproject.toml is). Is it possible to stick to this style in the editor? (I guess we should write a style guide listing those rules)

wizard/core/state.py
```diff
@@ -0,0 +1,144 @@
...
+        del self.distribution[key]
+    def add_record_set(self, record_set: RecordSet) -> None:
+        self.record_sets.append(record_set)
+    def update_record_set(self, key: int, record_set: RecordSet) -> None:
+        self.record_sets[key] = record_set
+    def remove_record_set(self, key: int) -> None:
+        del self.record_sets[key]
+
+    @classmethod
+    def from_canonical(cls: Type[T], dataset: mlc.Dataset) -> T:
```
marcenacp, line 97:
```
def from_canonical(cls, dataset: mlc.Dataset) -> Metadata:
```

Otherwise, `T` is a generic here.

You may also need:

```python
from __future__ import annotations
```

wizard/requirements.txt
```diff
@@ -4,3 +4,4 @@ pandas
 pytest
 requests
 streamlit
+mlcroissant
```
marcenacp, line 7: Alphabetical order. (Allows to easily locate packages while the requirements grow.)

wizard/views/load.py
```diff
@@ -13,14 +16,15 @@ def render_load():
...
-
+            file_cont = file.read()
+            newfile_name = Path.home() / ".cache" / "croissant" / "loaded_croissant"
+            os.makedirs(os.path.dirname(newfile_name), exist_ok=True)
+            with open(newfile_name, mode="wb+") as outfile:
+                outfile.write(file_cont)
+            dataset = mlc.Dataset(newfile_name)
+            st.session_state[Metadata] = Metadata.from_canonical(dataset)
+            set_form_step("Jump", CurrentStep.editor)
+            st.rerun()
```
marcenacp, line 27: Question: is `st.rerun()` safe?

wizard/core/state.py
```diff
@@ -0,0 +1,144 @@
+"""Streamlit session state.
```
marcenacp, line 1: The present file seems Streamlit-related. In `wizard/core`, the goal was to only put re-usable generic Python functions: like infering types, handling files, etc.

Is it possible to have those files isolated somewhere? You can certainly rename the folder, but I'd like to have the business logic isolated from the Streamlit logic (e.g., in case, we have to refactor with another web framework).

wizard/views/load.py
```diff
@@ -1,8 +1,11 @@
 import os
+from pathlib import Path
```
marcenacp, line 2: Can you use `epath.Path` as we do in mlcroissant?

```
from etils import epath

epath.Path(...)
```

marcenacp: `Path.home()` == `epath.Path("~").expanduser()`

wizard/views/record_sets.py
```diff
@@ -11,31 +12,29 @@
 
 
 def render_record_sets():
-    if not st.session_state[RecordSets]:
-        st.markdown("Provide files before.")
+    if len(st.session_state[Metadata].record_sets) == 0:
```
marcenacp, line 15: Prefer `if not list` (https://google.github.io/styleguide/pyguide.html#2144-decision).

wizard/state.py
marcenacp: Interesting. What shape/structure would you rather have?

wizard/state.py
marcenacp: Correct!

wizard/views/record_sets.py
drayuka: I understand that it's in the (google) style guide, but that is a terrible style/readability decision, it forces readers to figure out why it works (if they don't know), rather than relying on the plainly readable above.  Python styling/readability decisions (regardless of the source) seem to want to eschew readability to add "pythonicity" which functionally just means arcane decisions of a specific type (almost always arbitrary or involved in using language specific features, seemingly only to justify their continued existence).  Even the style guide you just linked has recommendations but no justifications for *why* this should be done, especially when it makes this flatly less readable. I generally have no problem running automated linters over code (as developer time spent is tiny). However, I would prefer not getting into these conversations as we're under time limitations that make such arbitrary style nits a tertiary concern at best, in my opinion.

drayuka: apologies if this was a bit of a rant, I've had python developers give "pythonic code" the same level of adherence as code structure or nominal functional code boundaries (things which do help readability, long term code health, and readability) and i'm always puzzled as to the value of writing code this way.  To date, none of them have been able to explain *why* it has value, in much the same way that golang developers can't explain why golang's "simplicity" is a positive trait.

marcenacp: Thanks for expressing your opinion and the detailed comments :)

The *why* justification of the rule is in the style guide I linked ([link](https://google.github.io/styleguide/pyguide.html#2142-pros)):

```
Conditions using Python booleans are easier to read and less error-prone. In most cases, they’re also faster.
```

This style guide is the one we've used so far, which is why I linked it. For example, reading through it can help you understand why the line used to be `if not st.session_state[RecordSets]:`.

I'd like us to stick to homogeneous rules within the code base. The Google style guide happens to be the style we picked since the beginning. As we start the editor from the existing codebase of mlcroissant, I guess it's just easier to stick with the existing style. :) WDYT?

wizard/views/load.py
drayuka: in what context? it just reruns the script like a UI interaction does, in this case that's required because just changing app state doesn't cause a re-render.

wizard/core/files.py
drayuka: I spent a couple of hours trying to figure this out (if you look at some of my commit messages, I lamented how non-simple python imports are, compared to say, javascript).  If you can explain how to avoid relative imports here, I'd be glad to hear it, but i suspect if I spend more brain power trying to figure out how python imports work/are sane I'll forget something else important, like how to speak more than 3 words in a row without staring into space, or how to talk without drooling.

wizard/core/state.py
drayuka: moved to utils instead, it's state related, but I did notice that core was streamlit free.

wizard/views/record_sets.py
drayuka: I understand the desire to follow the style guide, but I would be remiss if I didn't point out the folly of adhering to a style guide while a project deadline looms this close. We should talk about this on Thursday, but I'm a little concerned with the desire for quality over the desire for execution, and what that means for what we can accomplish before the end of November.  I'm not sure it's a great use of my time to devote this much effort into making each PR as high of a quality as you are clearly used to.   I would much prefer to generate a bit of tech debt/messy code now and get much more done feature wise, I certainly have no problem addressing this after the end of November (or even before, if we get most of our desired features done), even if I am no longer under contract, but don't feel like this is a great use of the time we have now.  In addition, it seems like some of these nits are getting in the way of discovering actual bugs, as I've noticed and caught a couple that have been completely missed.

marcenacp: I hope to achieve both quality and execution! Changing a line of code shouldn't take too much time. We can discuss on Thursday to see where we put the bar.

The point on catching bugs is a good point... From now on, I'll checkout the branch locally to have a closer look and hopefully catch more bugs. Thanks!

marcenacp (approved):

PR #317, opened 2023-11-08
Expose more nodes.* in init

ccl-core:
Update __init__.py to expose more mlcroissant.nodes attributes, also changed scripts in recipes/ and scripts/ to directly use them


marcenacp (approved):
Thanks!

PR #323, opened 2023-11-09
Proposal: new properties `ml:annotationOfField` and `ml:format` especially useful for bounding boxes.

marcenacp:
- `ml:annotationOfField` shows the field in the ml:RecordSet that is being annotated, e.g. in this example the bounding box is the annotation for an image.
- `ml:format` is the format of the bounding box, as per [Keras bounding box format](https://keras.io/api/keras_cv/bounding_box/formats/): `XYXY`, `XYWH`, etc.

Other types (like `sc:AudioObject`) will need additional properties.

### DIFF
--- a/datasets/coco2014-mini/data/captions_train2014.json
+++ b/datasets/coco2014-mini/data/captions_train2014.json
@@ -93,22 +93,22 @@
     ],
     "annotations": [
         {
-            "image_id": 318556,
+            "image_id": 57870,
             "id": 48,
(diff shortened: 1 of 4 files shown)


marcenacp:
Offline-discussions with Omar and Pierre:

- `format` should go in `extract.format` (new property).
- The relation object->label is represented by https://schema.org/additionalProperty on the object (here the image): `"annotatedBy": "label"`

PR #328, opened 2023-11-10
Use tabs for create/load dataset steps

aidazolic:
Replaces expanders with tabs on the create new / load existing configuration page.

### DIFF
--- a/wizard/views/overview.py
+++ b/wizard/views/overview.py
@@ -1,0 +1,2 @@
+def render_overview():
+    pass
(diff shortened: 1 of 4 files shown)


wizard/views/overview.py
```diff
@@ -0,0 +1,4 @@
+
+
+def render_overview():
```
marcenacp, line 3: If you rebase, you should be able to format using `black`.

marcenacp (approved):
LGTM, thanks!

Can you change the selector in the test (`[data-testid="stExpander"]`) with the new selector for tabs?

wizard/views/overview.py
aidazolic: Done

aidazolic:
> LGTM, thanks!
> 
> Can you change the selector in the test (`[data-testid="stExpander"]`) with the new selector for tabs?

I modified the tests.
Can you PTAL again? Thanks!

marcenacp (approved):
Awesome, thanks!

PR #341, opened 2023-11-13
Change Resources page UI

aidazolic:
Moves the list of already uploaded files up, and shortens the upload form.

### DIFF
--- a/wizard/views/files.py
+++ b/wizard/views/files.py
@@ -40,8 +40,16 @@ def _render_left_panel(files: list[Resource]) -> Resource | None:
         else:
             parent = None
         nodes.append({"name": name, "type": type, "parent": parent})
-    _render_upload_form()
-    name = render_tree(nodes)
+
+    name = None
+    with st.container():
+        st.subheader("Uploaded resources")
+        name = render_tree(nodes)
+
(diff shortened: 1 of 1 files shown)


aidazolic:
The updated UI:

Screenshot 2023-11-13 19 03 21

The "Add manually" button is also to be added, but in a separate PR.

marcenacp (approved):
Thanks!

benjelloun:
Cool! A couple of small suggestions:

- "Uploaded resources" should probably be just "Resources" since they may not be "uploaded", but provided as a URL or read from local disk ("uploading" to a locally running server is a bit strange).
- Similarly "Upload more files" should probably say "Add more ..." files? data? resources? Not sure what's the best terminology here.
- The big block of options for adding a file is only useful while adding a file. Maybe it makes more sense to have it in a modal dialog? This makes it harder to support drag and drop, but maybe that's not so important.
- Why is the content type drop down on the left and not on the right, with other attributes of the FileObject / FileSet?
- Where do you specify that a resource is a FileObject or a FileSet?
- Do we want to show more metadata in the list of files? Last update date, size, content type? Just in case they are useful for sorting.

PR #369, opened 2023-11-19
previous file load

### LINKED ISSUE #305: ability to load previously edited croissant files
I should be able to see and load up previously edited files from the croissant editor, this should probably be powered through cookies, possibly look at: https://discuss.streamlit.io/t/cookies-support-in-streamlit/16144

drayuka:
closes #305

### DIFF
--- a/editor/core/past_projects.py
+++ b/editor/core/past_projects.py
@@ -1,0 +1,62 @@
+import copy
+import dataclasses
+import os
+import pickle
+import time
+
+from etils import epath
+
+from .state import Metadata
+
+METADATA_DIRECTORY: str = epath.Path("~").expanduser() / ".cache" / "croissant"
+METADATA_FILE: str = METADATA_DIRECTORY / "previously_loaded"
+
+# super simple, second tuple is for some metadata around when this file was originally created/loaded
+projects: list[tuple[Metadata, int]] = []
+
+loaded: bool = False
+
+
+# inital load up of metadata file
+def load_past_projects() -> list[Metadata]:
+    global loaded
+    global projects
+    if not loaded:
+        os.makedirs(METADATA_DIRECTORY, exist_ok=True)
+        try:
+            open(METADATA_FILE, "x")
+        except:
+            pass
+        with open(METADATA_FILE, "rb+") as file:
+            try:
+                projects = pickle.loads(file.read())
+            except:
+                projects = []
+        if not projects:
+            projects = []
+        loaded = True
+    return projects
+
+
+# move loaded project to the front so we can see it
+def loaded_project(index: int) -> None:
+    # bring selected item to the front of the list
+    projects.insert(0, projects.pop(index))
+    save_projects()
+
+
+# run if a new croissant file is created, or if we load a new file
+def add_new_project(metadata: Metadata) -> None:
+    projects.insert(0, (metadata, time.time()))
+    # check to see if we have more than 10 previous projects, remove older ones
+    if len(projects) > 10:
+        projects.pop()
+    # save projects list back to the overall metadata file
+    save_projects()
+
+
+# save-as you go feature, first project is always the current one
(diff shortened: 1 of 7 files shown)


editor/core/past_projects.py
```diff
@@ -0,0 +1,63 @@
...
+
+# super simple, second tuple is for some metadata around when this file was originally created/loaded
+projects: list[tuple[Metadata, int]] = []
+
+loaded: bool = False
+
+
+# inital load up of metadata file
+def load_past_projects() -> list[Metadata]:
+    global loaded
```
marcenacp, line 23: I wouldn't use globals here, nor store projects in a specific data structure. You could rather have the filesystem represent the truth for all past datasets.

Advantages:
- Cleaner to order everything on disk in their own folder. Non finished datasets and finished datasets will have the same structure on disk.
- We'll be able to also save uploaded data.

I would propose the following structure:

```
$> ls -lrt ~/.croissant/cache/editor

|_ titanic-202309131455 # a fixed format concatenating name + creation date
  |_ .metadata.json  # temporary dump of Metadata (trailing point to indicate that it's private)
  |_ data  # later we will be able to add the data that was uploaded
|_ coco2014-202309132000
  |_ etc
```

So past projects would read from disk and render:

```
titanic - 2023/09/13 at 14:55
coco2014 - 2023/09/13 at 20:00
...
```

WDYT?

marcenacp (commented):
Thanks, Madelyn!

I proposed an alternative approach, which could allow us to go faster for the next features (e.g., saving local data, downloading the zipped dataset with its data, etc).

editor/core/past_projects.py
drayuka: a couple of problems:
1. you can't use names, since a new croissant file might not actually have one, ran into this, its a problem that's annoying to deal with, and having to change filenames as the user changes the name of a new croissant file (or an existing one) is a silly problem to try to get right given the amount of time we're working with.
2. Is this worth the time cost? the main output of this is that files that the user will never see look nicer on disk. If the goal is a complete dataset, going from Metadata->zip seems like the feature that solves that, not a different way of storing Metadata temporarily on disk (I understand how these are related, but combining these features increases complexity)
3. Kurt is wanting work done on the viewer and I'm going to devote some time to that this week after I get docker container builds working, So I'm going to be context switching a bit.

drayuka: note, I think this is a good way of doing this (and what I was trying some part of this weekend) and I picked the current solution because it just works, without any real problems that the current solution doesn't solve (besides being pre-prepared to be turned into a .zip archive)

marcenacp: > a new croissant file might not actually have one

Should the name only contain the timestamp in a first iteration?

> Is this worth the time cost?

With the current approach, how are you going to implement:
- Saving the data that was uploaded by the user? (Otherwise, it is lost at each refresh)
- Running the generation on the dataset (https://github.com/orgs/mlcommons/projects/29?pane=issue&itemId=44846512)?

> Kurt is wanting work done on the viewer

I thought we agreed viewer == editor == wizard to factorize the effort.

drayuka:
> Approving to unblock you, but I don't see the changes we spoke about in the comments and during the daily.

If you fire it up, it will now create the directories, I skipped on trying to maintain name-timestamp directories as it's complex and doesn't offer alot of value to users (zip archives can be named whatever) and just went for timestamp, but all metadata about the previous projects is now stored in the filesystem where
DIRECTORY_TIMESTAMP
   - metadata (last changed timestamp)
DIRECTORY_TIMESTAMP
   - metadata (last changed timestamp)
...
is the structure and both timestamps are used to display creation date/times and order items in the list so re-loading a previous item will push it to the front of the displayed list

from there, adding a feature in the files section to download items to the appropriate directories should be simple (but is not currently handled by this code)

PR #373, opened 2023-11-20
Save manually created resource in the config

aidazolic:
Adds the functionality to save a manually created resource. The ability to add data to this resource (e.g. `Record Sets`) will be added separately. 

See the demo [here](http://screencast/cast/NjI4OTM3NDQ4NzUxMTA0MHwzMTk2NWFlNy00NQ)

### DIFF
--- a/editor/core/files.py
+++ b/editor/core/files.py
@@ -9,6 +9,11 @@ import requests
 
 from .names import find_unique_name
 from .state import FileObject
+from .state import FileSet
+
+FILE_OBJECT = "File object"
+FILE_SET = "File set"
+RESOURCE_TYPES = [FILE_OBJECT, FILE_SET]
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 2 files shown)


editor/core/files.py
```diff
@@ -124,3 +125,24 @@ def file_from_upload(
         sha256=sha256,
         df=df,
     )
+
+
+def file_from_form(
+    file_type: FileType, type: str, name, description, sha256: str, names: set[str]
+) -> FileObject | FileSet:
+    """Creates a file based on manually added fields."""
+    if type == "File object":
```
marcenacp, line 134: Is there a way to not use strings here?

editor/views/files.py
```diff
@@ -115,10 +120,26 @@ def handle_on_click():
...
+                needs_sha256 = resource_type == FILE_OBJECT
+
+                name = st.session_state[_MANUAL_NAME_KEY]
+                description = st.session_state[_MANUAL_DESCRIPTION_KEY]
+                sha256 = st.session_state[_MANUAL_SHA256_KEY] if needs_sha256 else None
+
+                if not name or (needs_sha256 and not sha256):
+                    # Some required fields are empty.
+                    st.toast(
+                        "Please, import either a local file, provide a download URL, or"
```
marcenacp, line 134: Is the error message relevant to the error? `Please, fill in all required fields: ...`?

marcenacp (approved):
Thanks!

editor/core/files.py
aidazolic: Done, replaced with constants but I can clean this up a bit more in a dedicated PR

editor/views/files.py
aidazolic: Modified the error message. Basically we show the error if neither uploaded file, nor URL, nor the form with the manual input is filled in so that we can actually add a Resource. Probably would be better to have a dedicated error message for each "option".

Also, do we want to maybe not clear the form if it wasn't handled successfully? This way if the user e.g. inputs name, description and clicks "Upload" we don't delete everything but they can just add the SHA256.

PR #391, opened 2023-11-23
Deploy to Hugging Face Spaces.

marcenacp:
- Try/catch pickling in case it creates errors.
- Deploy to Hugging Face Spaces.


ccl-core (approved):
Thanks!

PR #408, opened 2023-11-29
Pre-visualize the first rows.

marcenacp:


editor/views/overview.py
```diff
@@ -47,29 +47,51 @@ def render_overview():
...
-                st.markdown("##### Errors:")
-                st.write(str(exception))
+                warning += "**Errors**\n"
+                warning += f"{str(exception)}\n"
+            if warning:
+                st.warning(warning, icon="⚠️")
+            else:
+                st.success("No validation issues detected!", icon="✅")
+        st.info(
+            """Croissant files are composed of three layers:
```
ccl-core, line 82: Nit, Might be cleaner to keep those long text variables separated, where they are slightly easier to read?

editor/views/record_sets.py
```diff
@@ -29,6 +29,29 @@
 ]
 
 
+def _generate_data(record_set: RecordSet) -> pd.DataFrame | None:
+    """Generates the first 3 records."""
```
ccl-core, line 33: Is there any specific reason why 3? Would it make sense to parametrize it?

ccl-core (approved):
Thanks!

By the way, we are not unit-testing the editor?

marcenacp:
> By the way, we are not unit-testing the editor?

No, we favored end-to-end tests

editor/views/overview.py
marcenacp: Done.

editor/views/record_sets.py
marcenacp: Done.

PR #410, opened 2023-11-29
Add version property to datasets.

ccl-core:


ccl-core:
A question here would be whether we'd like to force the version to have any specific format (e.g. "1.0" is accepted but "1" no, or even "1.0.0" not).

I'd be in favor or keeping this very flexible for now, and not to add any extra check; but I would be happy to add some format check in `metadata.py`.

ccl-core:
Note: https://schema.org/version in schema.org allows a version to be of types: "Text" and "Number".

I am not sure whether we'd like to support "Number" in mlcroissant or not.
This would work if we allow users to only specify versions as `MAJOR.MINOR`, but we might get into problem if users use integers and then select `1.0.0` -- WDYT? I'd be happy to add support for `floats` the version typing.

marcenacp:
Starting with strings (`"1.0"`) SGTM. We could certainly add checks, I think we'll need some checks anyway in the future

datasets/huggingface-c4/metadata.json
```diff
@@ -37,6 +37,7 @@
   "description": "A colossal, cleaned version of Common Crawl's web crawl corpus.\n\nBased on Common Crawl dataset: \"https://commoncrawl.org\".\n\nThis is the processed version of Google's C4 dataset by AllenAI.\n",
   "citation": "\n@article{2019t5,\n    author = {Colin Raffel and Noam Shazeer and Adam Roberts and Katherine Lee and Sharan Narang and Michael Matena and Yanqi Zhou and Wei Li and Peter J. Liu},\n    title = {Exploring the Limits of Transfer Learning with a Unified Text-to-Text Transformer},\n    journal = {arXiv e-prints},\n    year = {2019},\n    archivePrefix = {arXiv},\n    eprint = {1910.10683},\n}\n",
   "url": "https://huggingface.co/datasets/c4",
+  "version": "0.0.0",
```
marcenacp, line 40: Should this be `"0.0"` or is the patch intended here?

marcenacp (approved):

datasets/huggingface-c4/metadata.json
ccl-core: If we import from HF, then the versioning that they use is indeed `MAJOR.MINOR.PATCH` so I am maintaining it to avoid confusion

PR #413, opened 2023-11-30
Support more licenses from https://huggingface.co/docs/hub/repositori…

marcenacp:
…es-licenses.

### DIFF
--- a/editor/views/metadata.py
+++ b/editor/views/metadata.py
@@ -11,6 +11,7 @@ from events.metadata import MetadataEvent
 
 
 def render_metadata():
+    """Renders the `Metadata` view."""
     metadata = st.session_state[Metadata]
     index = find_license_index(metadata.license)
     key = "metadata-license"
--- a/editor/views/metadata.py
+++ b/editor/views/metadata.py
@@ -3,40 +3,25 @@ import enum
 import streamlit as st
 
 from core.state import Metadata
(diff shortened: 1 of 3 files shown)


editor/views/metadata.py
```diff
@@ -3,40 +3,25 @@
...
-    "CDLA-Sharing-1.0",
-    "ODC-ODbL",
-    "CC BY-NC",
-    "CC BY-ND",
-    "CC BY-NC-SA",
-    "CC BY-NC-ND",
-]
-
 
 def render_metadata():
```
ccl-core, line 13: Any chance we can add a docstring?

editor/views/metadata.py
```diff
@@ -3,40 +3,25 @@
...
     metadata = st.session_state[Metadata]
-    try:
-        index = licenses.index(metadata.license)
-    except ValueError:
-        index = None
+    index = find_license_index(metadata.license)
```
ccl-core, line 15: What happens if the value is None?

ccl-core (approved):
Thanks!

editor/views/metadata.py
marcenacp: `None` is a supported value :)

editor/views/metadata.py
marcenacp: Done.

PR #415, opened 2023-12-01
Print "Text" instead of "https://schema.org/Text".

marcenacp:
Fixes: https://github.com/mlcommons/croissant/issues/401


ccl-core (approved):
Nice!

PR #420, opened 2023-12-04
Handle local data.

marcenacp:
Add a warning when the local data does not exist.
Save the data locally.
Encoding formats are in a select box.
Download files in data/ for Titanic
Use python-magic to guess the file type.
Add info on RecordSets.
Allow to re-trigger the dataset generation.
Gamify the experience with a percentage of completion.

### DIFF
--- a/editor/core/files.py
+++ b/editor/core/files.py
@@ -4,6 +4,7 @@ import io
 import tempfile
 
 from etils import epath
+import magic
 import pandas as pd
 import requests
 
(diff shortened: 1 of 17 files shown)


editor/core/files.py
```diff
@@ -84,58 +132,91 @@ def download_file(url: str, file_path: epath.Path):
...
-        return pd.read_json(file)
+        df = pd.read_json(file)
     elif file_type == FileTypes.JSONL:
-        return pd.read_json(file, lines=True)
+        df = pd.read_json(file, lines=True)
     elif file_type == FileTypes.PARQUET:
-        return pd.read_parquet(file)
+        df = pd.read_parquet(file)
     else:
         raise NotImplementedError()
```
pierrot0, line 145: Can we add a message, something like "File format XXX no supported"? That's less cryptic, and people know what to ask for.

pierrot0 (approved):
Thanks!

editor/core/files.py
marcenacp: Done in another PR (https://github.com/mlcommons/croissant/pull/423/commits/1d2ccf93b23704a522610c68e02a1b165b849298).

PR #429, opened 2023-12-06
Do not cache user in backend.

marcenacp:
Reason: In Streamlit, cache is not handled per session, but cache is common to all users.


ccl-core (approved):
Thanks!

PR #439, opened 2023-12-06
Add `conformsTo` field.

ccl-core:


datasets/bigcode-the-stack/metadata.json
```diff
@@ -37,6 +39,7 @@
   },
   "@type": "sc:Dataset",
   "name": "bigcode-the-stack",
+  "conformsTo": "http://mlcommons.org/croissant/1.0",
```
marcenacp, line 42: Question (for all files): should it be `https` instead of `http`? (Like `ml`)

marcenacp (approved):
Thanks!

PR #454, opened 2024-01-22
Make mlcroissant compatible with older versions of rdflib/rdflib-jsonld.

marcenacp:


ccl-core (commented):
That's great, thank you!!!

pierrot0 (approved):

PR #458, opened 2024-01-22
Define multiple record sets deriving from the same node

ccl-core:
This is needed in order to have multiple RecordSets (deriving e.g. from different ReadFields operations) which derive from the same Read operation. As an example, refer to the coco2014-mini dataset below.

image (7)


python/mlcroissant/mlcroissant/_src/operation_graph/operations/parse_json.py
```diff
@@ -7,10 +7,15 @@
...
 
-def parse_json_content(json: Json, fields: tuple[Field, ...]) -> pd.DataFrame:
+def parse_json_content(
+    json: Json, fields: tuple[Field, ...], record_set: str | None = None
+) -> pd.DataFrame:
     """Parsed all JSONs defined in the fields of RecordSet and outputs a pd.DF."""
     series = {}
     for field in fields:
+        # Only select fields relevant to the requested record_set (if given).
+        if record_set and not field.uid.startswith(record_set):
```
marcenacp, line 17: Here can you filter at the field level (`fields`) rather than adding another parameter (`record_set`)?

marcenacp (changes requested):
Thanks, LGTM! I think there's possibly a simplification because `Read` operations should not have to know about the RecordSets (children)

python/mlcroissant/mlcroissant/_src/operation_graph/operations/parse_json.py
ccl-core: Good point, changed. WDYT?

python/mlcroissant/mlcroissant/_src/operation_graph/base_operation.py
```diff
@@ -25,18 +26,28 @@ def last_operations(self, node: Node, only_leaf=False) -> list[Operation]:
...
             # node1 is predecessor of node2 iff node2 is in the descendants of node1.
             ancestors = nx.ancestors(node2.ctx.graph, node2)
             if isinstance(node1, RecordSet):
                 # If node1 is a RecordSet, we have to inspect its fields.
-                return any(is_ancestor(field1, node2) for field1 in node1.fields)
+                return any(
+                    is_ancestor(field1, node2, ancestor_leaf) for field1 in node1.fields
+                )
+
+            if hasattr(ancestor_leaf, "is_read_operation") and isinstance(node2, Field):
```
marcenacp, line 38: If `Read` overwrote an `is_ancestor` method, we could avoid this workaround :)

python/mlcroissant/mlcroissant/_src/operation_graph/graph.py
```diff
@@ -57,6 +57,26 @@ def _add_operations_for_field_with_source(
...
+def _get_recordsets_for_fields(file_obejct: FileObject) -> dict[str, Field]:
+    """Gets dictionary of RecordSet and their respective Fields generated from a node.
+
+    Args:
+        file_obejct: A FileObject.
+
+    Returns:
+        A dictionary of RecordSet names, and the Fields relative tho that RecordSet.
+    """
+    recordset_to_fields = {}
```
marcenacp, line 69: Use a defaultdict?

marcenacp:
Closing this PR as it is now possible to define multiple RecordSets deriving from the same node.

PR #462, opened 2024-01-25
Support mapping resources to files on disk (`--mapping`).

marcenacp:


python/mlcroissant/mlcroissant/scripts/load.py
```diff
@@ -42,6 +43,15 @@
     "Whether to update the JSONL output test files.",
 )
 
+flags.DEFINE_string(
+    "mapping",
+    None,
+    "Mapping filename->filepath as a Python dict[str, str] to handle manual downloads."
+    " If `document.csv` is the FileObject and you downloaded it to"
```
pierrot0, line 50: This is good, but maybe I would also offer a flag to just specify a directory where mlcroissant library can find all files listed in the Croissant config. Maybe the mapping file can be re-used with some generic mapping like '{"(.+)": "~/Downloads/data/{$1}"', but that might not be the most user-friendly way.

pierrot0 (approved):

python/mlcroissant/mlcroissant/scripts/load.py
marcenacp: Do we need both flags? For instance, if you want to manually download only one resource, but other resources are to be taken from the web?

marcenacp: We spoke offline with @pierrot0 and decided to go with the current implementation. As far as supporting a directory is concerned, we'll wait for a user to have a specific need for that option.

PR #465, opened 2024-01-26
Load JSON-LD from URL in addition to load from a file.

marcenacp:
This PR also fixes the GitHub Actions (some of them are broken and don't trigger anymore).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,8 +205,9 @@ jobs:
     defaults:
       run:
         working-directory: ./python/mlcroissant
-    matrix:
-      versions: ["0.8", "1.0"]
+    strategy:
+      matrix:
+        version: ['0.8', '1.0']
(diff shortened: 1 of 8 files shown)


goeffthomas:
Saw the notification for this and it looks amazing! When do you think we could get this pushed out for download via `pip`? I'm making an example Kaggle notebook to show folks how to load a dataset and it's a minor papercut to have to do the extra steps of fetching/storing the JSON to use when I just want what's at the URL.

marcenacp:
> When do you think we could get this pushed out for download via pip?

I hope on Mon, Jan 29, as soon as it gets approved. Glad that you like it. In the future we can have an even more simple way to load Kaggle datasets (like `mlc.KaggleDataset(name, **kwargs)`) so that users don't have to know about the details of the exact endpoint

pierrot0 (approved):

PR #479, opened 2024-02-05
Migrate parameters that don't exist in schema.org.

marcenacp:


python/mlcroissant/mlcroissant/_src/core/rdf_test.py
```diff
@@ -1,16 +1,18 @@
 """Tests for RDF."""
 
+from mlcroissant._src.core.context import Context
 from mlcroissant._src.core.rdf import Rdf
 
 
 def test_shorten_value():
     rdf = Rdf.from_json(
+        ctx=Context(),
```
ccl-core, line 9: Would it make sense to add a test for when ctx is not empty?

ccl-core (approved):
Thanks Pierre!

python/mlcroissant/mlcroissant/_src/core/rdf_test.py
marcenacp: Good idea, thanks!

PR #481, opened 2024-02-05
Install from source from GITHUB_HEAD_REF and default to main.

marcenacp:


ccl-core (approved):
Thanks, that's really useful!

PR #484, opened 2024-02-06
Fix dev container problems

ccl-core:


.devcontainer/postCreateCommand.sh
```diff
@@ -1,7 +1,20 @@
 #!/bin/bash
 
-sudo apt-get update
-sudo apt-get install -y libgraphviz-dev
+sudo apt-get install -y libgraphviz-dev git-lfs
 
-# Install dev packages.
-pip3 install python/mlcroissant/.[dev]
+# Install dependencies except mlcroissant itself
+pip install absl-py \
```
marcenacp, line 6: Not possible to do something like `pip install .[dev]`? (No idea, would just be easier to maintain)

python/mlcroissant/pyproject.toml
```diff
@@ -35,12 +35,15 @@ readme = "README.md"
...
 dev = [
-  "black==23.11.0",
-  "datasets",
-  "flake8-docstrings",
+  "mlcroissant[devdeps]",
   "mlcroissant[git]",
   "mlcroissant[image]",
   "mlcroissant[parquet]",
+]
+devdeps = [
```
marcenacp, line 43: Why do you need to distinguish between dev and devdeps and what distinction do you make?

marcenacp (approved):

PR #486, opened 2024-02-06
Memoize jsonpath_rw.parse for better performances.

marcenacp:
Before: 675.2771615982056 seconds
After: 11.278749465942383 seconds


ccl-core (approved):
Makes a lot of sense, thanks!

PR #491, opened 2024-02-06
Fix flaky test

aidazolic:
The uploadCsv test was sometimes failing to verify that the Description field was modified.

One of the possible issues was not finding the Description field to type into and check, as it wouldn't be visible in the current view. This was also made more likely due to warnings to migrate usages of  `st.expermental_get_query_params` 
 and `st.expermental_set_query_params` which are now replaced with `st.query_params`.   
In the test itself, now waits after clilcking the "Edit fields" button, so that the necessary input fields appear, instead of waiting after typing and modifies the element lookup.

### DIFF
--- a/editor/app.py
+++ b/editor/app.py
@@ -21,7 +21,7 @@ init_state()
 user = get_user()
 
 if OAUTH_CLIENT_ID and not user:
-    query_params = st.experimental_get_query_params()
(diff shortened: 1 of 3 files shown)


marcenacp (approved):
Good catch, let's merge to see if it works. Thanks!

PR #492, opened 2024-02-06
AAU, I can override CROISSANT_CACHE with an environment variable or a script parameter

marcenacp:
Fixes: https://github.com/mlcommons/croissant/issues/453


ccl-core (approved):
Thanks!

Issue #501, opened 2024-02-08
Generation errors in `bigcode-the-stack`, `coco2014-mini` and `simple-dataset`


marcenacp:
Reported by @St3V0Bay

marcenacp:
`simple-dataset`: to be removed as unused

ccl-core:
`simple-dataset` and `bigcode-the-stack` shouldn't create more problems now.

For `coco2014-mini`, #510 proposes an ad-hoc fix, but if time allows, it would make sense to have something less hacky in place before the launch.

PR #503, opened 2024-02-08
Delete simple dataset as it is unused and it causes problems.

ccl-core:


ccl-core:
Addresses #501

marcenacp (changes requested):
Is there also a 0.8 version?

ccl-core:
There is, I deleted it too. Sorry, I thought we wanted to delete only the 1.0

marcenacp (approved):

PR #505, opened 2024-02-09
Fix query params

aidazolic:
`query_params` is not a `dict` so usage is a bit different from the return values of `st.experimental_get_query_params`, see [documentation](https://docs.streamlit.io/library/api-reference/utilities/st.query_params)


marcenacp (approved):
Thanks!

PR #507, opened 2024-02-12
subset of updated checksums

St3V0Bay:


datasets/0.8/fashion-mnist/metadata.json
```diff
@@ -96,4 +96,4 @@
       ]
     }
   ]
-}
+}
```
marcenacp, line 99: Can you keep the end-of-line?

marcenacp (changes requested):
Thanks a lot for the PR!

datasets/0.8/fashion-mnist/metadata.json
St3V0Bay: sure sure. that's a bit a tricky one. VSC removes this automatically when saving the file unless I set the following 
Screenshot 2024-02-12 at 20 05 54

marcenacp: Yes! Can you please set this option?

marcenacp (approved):

PR #515, opened 2024-02-14
Support both old and new query_params

aidazolic:
Since Streamlit version 1.30.0., all usages of [`st.experimental_get_query_params()`](https://docs.streamlit.io/library/api-reference/utilities/st.experimental_get_query_params) and [`st.experimental_set_query_params()`](https://docs.streamlit.io/library/api-reference/utilities/st.experimental_set_query_params) should be replaced by [`st.query_params`](https://docs.streamlit.io/library/api-reference/utilities/st.query_params).

However, in older version, using `st.query_params` gives an attribute not found error, therefore modifying the editor to support both the old and the new API.

### DIFF
--- a/editor/core/query_params.py
+++ b/editor/core/query_params.py
@@ -15,20 +15,44 @@ class QueryParams:
     OPEN_RECORD_SET = "recordSet"
 
 
+def _has_query_params() -> bool:
+    """Returns whether `query_params` is available, which depends on the streamlit
+    version. If not, st.experimental_get_query_params() and
+    st.experimental_set_query_params() should be used."""
+    return hasattr(st, "query_params")
+
+
 def _get_query_param(name: str) -> str | None:
     """Gets query param with the name `name`."""
-    param = st.query_params.get_all(name)
+    if _has_query_params():
+        param = st.query_params.get_all(name)
+    else:
+        params = st.experimental_get_query_params()
+        if not name in params:
+            return None
+        param = params[name]
+
     if isinstance(param, list) and len(param) > 0:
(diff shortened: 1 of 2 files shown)


editor/core/query_params.py
```diff
@@ -15,20 +15,40 @@ class QueryParams:
...
+        if params.get_all(param) == [new_value]:
+            # The value already exists in the query params.
+            return
+        params[param] = new_value
+    else:
+        params = st.experimental_get_query_params()
+        if params.get(param) == [new_value]:
+            # The value already exists in the query params.
+            return
+        new_params = {k: v for k, v in params.items() if k != param}
```
marcenacp, line 49:
```
del params[param]
```

But why do you need to copy like this the dict to then reaffect the new value. Why not just `new_params[param] = new_value`?

marcenacp (approved):
Thanks for the quick fix! Did you test on a previous version of Streamlit?

editor/core/query_params.py
aidazolic: This is how we had it before. I think because otherwise the `set` would erase existing values.

aidazolic:
> Thanks for the quick fix! Did you test on a previous version of Streamlit?

Yes, but I'm not sure I tested all the possible scenarios. 😃
Basically I installed Streamlit 1.29.0 and played around the editor, and the same with the new 1.31.0. 

There are some problems in the CI that I need to check though.

aidazolic:
Seems that all tests are passing now, I'll merge this and we can see if all the issues are fixed.

PR #516, opened 2024-02-14
Flores200 dataset and notebook refresh

mkuchnik:
The Flores 200 dataset was missing a version field, causing warnings. The notebook also refers to old paths that no longer exist. This patch fixes these issues and tries to align the style of the notebook with the other notebooks in the directory.


mkuchnik:
Closing for #517

PR #517, opened 2024-02-14
Flores200 dataset and notebook refresh v2

mkuchnik:
The Flores 200 dataset was missing a version field, causing warnings. The notebook also refers to old paths that no longer exist. This patch fixes these issues and tries to align the style of the notebook with the other notebooks in the directory.

Opening another PR on top of #516 to test if CI depends on a native branch.


mkuchnik:
CI should now (partially) check the FLORES notebook, which can help with regressions.

marcenacp (approved):

PR #527, opened 2024-02-16
Pass RDF `@id` to nodes as uuid.

marcenacp:
- Start deprecating `uid` to be replaced by `uuid`.
- Better hashing (as a consequence I discovered some issues in some datasets).

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/core/uuid.py
+++ b/python/mlcroissant/mlcroissant/_src/core/uuid.py
@@ -6,7 +6,7 @@ from mlcroissant._src.core.types import Json
 
 
 def generate_uid() -> str:
-    """Generates a random UUID of version 4."""
+    """Generates a UUID of version 4 because it's random and simple."""
     return str(uuid.uuid4())
 
 
--- a/python/mlcroissant/mlcroissant/_src/core/uuid.py
+++ b/python/mlcroissant/mlcroissant/_src/core/uuid.py
@@ -1,0 +1,18 @@
+"""Module to manipulate UUID."""
+
+import uuid
+
+from mlcroissant._src.core.types import Json
+
+
+def generate_uid() -> str:
+    """Generates a random UUID of version 4."""
+    return str(uuid.uuid4())
+
+
+def uuid_from_jsonld(jsonld: Json) -> str:
+    """Retrieves uuid from a JSON-LD fragment."""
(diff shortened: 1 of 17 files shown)


python/mlcroissant/mlcroissant/_src/core/uuid.py
```diff
@@ -0,0 +1,18 @@
+"""Module to manipulate UUID."""
+
+import uuid
+
+from mlcroissant._src.core.types import Json
+
+
+def generate_uid() -> str:
+    """Generates a random UUID of version 4."""
```
ccl-core, line 9: Maybe you could comment on why we choose version 4?

python/mlcroissant/mlcroissant/_src/datasets.py
```diff
@@ -114,15 +117,16 @@ def __iter__(self):
...
         if self.debug:
             graphs_utils.pretty_print_graph(operations)
         # Downloads can be parallelized, so we execute them in priority.
         execute_downloads(operations)
         # We can stream the dataset iff the operation graph is a path graph (meaning
         # that all operations lie on a single straight line, i.e. have an
         # in-degree of 0 or 1. That means that the operation graph is a single line
-        # (without external joins for example).
+        # (without external joins for example). We only consider the operations that
+        # are useful to produce the `ReadFields(record_set)`.
```
ccl-core, line 129: > We only consider the operations that are useful to produce the `ReadFields(record_set)`

This depends on _filter_interesting_operations() no?
If yes, then maybe this comment could be put above?

ccl-core (approved):
Thanks! I left a few NITs (mainly to help future readers to understand the technicalities)

python/mlcroissant/mlcroissant/_src/core/uuid.py
marcenacp: Done.

marcenacp:
@mkuchnik The notebook `recipes/flores200_datapipes.ipynb` fails. It seems a part of the code still uses an older version of mlcroissant, because it complains that the field I just introduced (`uuid`) doesn't exist. Also, the notebook runs in local. Do you see why it could fail? Thanks.

marcenacp:
I ended up implementing a simple `__deepcopy__` for all nodes and I created the issue https://github.com/mlcommons/croissant/issues/531 to track if there are any side effects.

PR #537, opened 2024-02-20
Allow to load RecordSets with inline data only.

marcenacp:
Fixes https://github.com/mlcommons/croissant/issues/501.

Full operation graph:

image

Operation graph for the RecordSet:

image


ccl-core:
Addresses #501

Issue #541, opened 2024-02-20
MIME type for TSV


mkuchnik:
The current MIME type for tab separated values is "text/tsv" to mirror "text/csv", though I now see that "text/tab-separated-values" is correct:
https://www.iana.org/assignments/media-types/text/tab-separated-values

References to "text/tsv" should maybe be migrated to "text/tab-separated-values".

PR #542, opened 2024-02-20
Add TSV to editor

mkuchnik:
Add tab separated values to the Croissant editor.

Tab separated values are used for some datasets (e.g., FLORES 200). See #541 for related issue.

### DIFF
--- a/editor/core/files.py
+++ b/editor/core/files.py
@@ -54,7 +54,7 @@ class FileTypes:
     )
     TXT = FileType(
         name="Text",
-        encoding_format="plain/text",
+        encoding_format="text/plain",
         extensions=["txt"],
     )
     ZIP = FileType(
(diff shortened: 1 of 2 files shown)


mkuchnik:
@marcenacp The MIME type inference is a bit flaky.

For example, the TSV from FLORES 200 is classified as "text/plain". The original MIME type was typo'd, so the user would see a match on `None` rather than `NotImplementedError: File type FileType(name='Text', encoding_format='text/plain', extensions=['txt']) is not supported. Please, open an issue on GitHub: https://github.com/mlcommons/croissant/issues/new`. The first commit addresses the mismatch to "text/plain" and the second converts "text/plain" to TSV. For the matching on TSV, I added a few additional heuristics to utilize extension information to promote plaintext to CSV/TSV. I also added some extra tests, though they don't exercise the error path FLORES 200 does when uploaded.

marcenacp (approved):
Thanks!

PR #555, opened 2024-02-23
Add ImageNet example

mkuchnik:
ImageNet is a common workload for benchmarking. This PR adds a recipe for running ImageNet in PyTorch with a Croissant loader, which can be useful for both system and ML characterization.

Current WIP status is:
* ImageFolder to Croissant prototype.
* Functional equivalence of Croissant loader with validation loader.

TODO:
* Add training loader with shuffling.
* Optimize system performance (e.g., startup for validation loader is ~30 seconds compared to near instant).


mkuchnik:
@marcenacp Thoughts on the structure so far? I'm not sure if there is a better way to read a local directory---I essentially emulate a zip/tar by creating a no-op "extract" function. I plan to simplify this further to avoid the symbolic links. I am also wondering if you have any thoughts on how to add shuffling/sharding of rows in `mlcroissant`, which would help with various node and worker-level parallelism strategies.

python/mlcroissant/mlcroissant/_src/operation_graph/operations/extract.py
```diff
@@ -20,15 +20,27 @@
...
+        or encoding_format == EncodingFormat.ZIP
+        or encoding_format == EncodingFormat.DIR
     )
 
 
+def _is_directory(source: epath.Path) -> bool:
+    return source.exists() and source.is_dir()
+
+
+def _soft_link(source: epath.Path, target: epath.Path) -> None:
```
marcenacp, line 33: Why do you need this function if `_soft_link == os.symlink`?

python/mlcroissant/examples/imagenet/main.py
```diff
@@ -0,0 +1,771 @@
+"""ImageNet training in PyTorch.
+
+From:
```
marcenacp, line 3: Do we need to copy this from pytorch into mlcroissant? Could you alternatively refer the user to https://github.com/pytorch/examples/blob/2d725b6ab255e05c55e0b08925f06f171aaedc0c/imagenet/main.py?

python/mlcroissant/mlcroissant/_src/core/constants.py
```diff
@@ -161,6 +161,7 @@ class EncodingFormat:
     """
 
     CSV = "text/csv"
+    DIR = "application/x-dir"
```
marcenacp, line 164: What's the difference with https://github.com/mlcommons/croissant/blob/e532c4c17d97ee660f40f2c21ac12b66ca0e990f/datasets/1.0/recipes/read_from_directory.json#L51-L59?

marcenacp (changes requested):

marcenacp:
@mkuchnik Sorry for the late review... I think it's super interesting and I left a few comments:

- There is already a way to import files from a specific folder (`data/`). I linked to an example.
- Maybe it's confusing to copy a whole Python file from pytorch. Can you give more instructions? Maybe it's worth having a Jupyter Notebook recipe with an end-to-end example on ImageNet!

PR #556, opened 2024-02-23
Add `dateCreated`, `dateModified`, `datePublished`.

marcenacp:


marcenacp:
(Tests are green because OpenML is down.)

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/metadata.py
```diff
@@ -237,6 +244,19 @@ def validate_license(self) -> list[str] | None:
             self.add_error(f"License should be a list of str. Got: {license}")
             return None
 
+    def validate_date(self, date: Any) -> datetime.datetime | None:
```
ccl-core, line 247: If time allows maybe add a parametrizxed test?

ccl-core (approved):
Thanks!

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/metadata.py
marcenacp: Done.

I did it in `test_from_jsonld` because `from_str_to_datetime` is already parametrized with various inputs. Thanks!

PR #558, opened 2024-02-23
Add https://schema.org/publisher.

marcenacp:


ccl-core (approved):
Nice! Any chance you can add some creators to one of our datasets to have an example for the users? Even one of the toy examples

PR #562, opened 2024-02-23
Respect the cardinality from the Croissant 1.0 specs.

marcenacp:


python/mlcroissant/mlcroissant/_src/core/json_ld.py
```diff
@@ -114,6 +114,20 @@ def unbox_singleton_list(d: Any):
     return d
 
 
+def box_singleton_list(element: Any) -> list[Any] | None:
+    """Returns the elemtn always boxed inside a list.
```
ccl-core, line 118: element

inside a list -> inside of a list

Can you also add an example of such property? I assume things such as containedIn woulf fix into this box?

python/mlcroissant/mlcroissant/_src/operation_graph/operations/filter.py
```diff
@@ -13,6 +15,23 @@
 from mlcroissant._src.structure_graph.nodes.file_set import FileSet
 
 
+@functools.cache
+def _memoized_regex(pattern: str) -> re.Pattern:
+    regex = fnmatch.translate(pattern)
+    return re.compile(regex)
+
+
+def match(patterns: list[str] | None, path: pathlib.PurePath) -> bool:
```
ccl-core, line 24: matches_path ?Maybe use a different name for this function? Match is a bit general maybe.

ccl-core (approved):
Nice!

python/mlcroissant/mlcroissant/_src/operation_graph/operations/filter.py
marcenacp: Done.

python/mlcroissant/mlcroissant/_src/core/json_ld.py
marcenacp: Done. The docstring is better now, thanks!

PR #569, opened 2024-02-27
First commit of Uuid class

ccl-core:


python/mlcroissant/mlcroissant/_src/core/uuid.py
```diff
@@ -1,43 +1,53 @@
...
 from mlcroissant._src.core.context import Context
 from mlcroissant._src.core.types import Json
 
 
+@dataclasses.dataclass
+class Uuid:
+    """Unique identifiers used in Croissant."""
+
+    uuid: None | str
+    ctx: Context
```
marcenacp, line 16: ctx as a first argument?

python/mlcroissant/mlcroissant/_src/structure_graph/base_node.py
```diff
@@ -38,12 +39,14 @@ class Node(abc.ABC):
     """
 
     ctx: Context = dataclasses.field(default_factory=Context)
-    id: str = dataclasses.field(default_factory=generate_uuid)
+    id: str | Uuid = dataclasses.field(default_factory=generate_uuid)
```
marcenacp, line 42: `id: Uuid`

We should try to avoid compound types as much as possible

python/mlcroissant/mlcroissant/_src/core/uuid.py
```diff
@@ -1,43 +1,53 @@
...
+
+    def __post_init__(self):
+        """Cleanup uuid."""
+        self.uuid = self.to_jsonld()
+
+    @classmethod
+    def from_jsonld(cls, jsonld: Json | None, ctx: Context) -> "Uuid":
```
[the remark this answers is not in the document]
ccl-core: Done.

python/mlcroissant/mlcroissant/_src/core/uuid.py
ccl-core: Done.

python/mlcroissant/mlcroissant/_src/structure_graph/base_node.py
ccl-core: I agree, but this is actually handy internally -- e.g. in metadata.py:390 we would need to first convert the jsonld to a string, then use it as input for id. Now it is more compact...

PR #575, opened 2024-02-29
[Proposal] Make it easier to extend Croissant, add new fields in mlcroissant and document them.

marcenacp:
This PR primarily aims at facilitating the extensibility of Croissant in mlcroissant. When extending Croissant with the RAI fields (PR https://github.com/mlcommons/croissant/pull/553), we found out it was complex to add new fields: code has to be added at several place and it's not trivial to test. This PR would solve this issue by having all property definitions next to the class itself using the [`dataclasses`](https://docs.python.org/3/library/dataclasses.html) paradigm.

I started in this PR with the example of `FileObject`:

```python
class FileObject(Node):
    """Nodes to describe a dataset FileObject (distribution)."""

    content_url: str | None = jsonld_field(
        default=None,
        description="Actual bytes of the media object, for example the image file or video file.",
        input_types=[SDO.URL],
        url=SDO.contentUrl,
    )
```

Another potentially desirable consequence of this refacto is that we could generate schema.org-like documentation **as supported/implemented by mlcroissant**:
image

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/core/dataclasses.py
+++ b/python/mlcroissant/mlcroissant/_src/core/dataclasses.py
@@ -23,7 +23,7 @@ class JsonldField(dataclasses.Field):
         input_types: list[Any],
         to_jsonld: Callable[[Context, Json], Any] | None,
         required: bool,
-        url: term.URIRef | Callable[[Context], term.URIRef] | None,
+        url: term.URIRef | Callable[[Context], term.URIRef],
     ):
         """Sets all args and kwargs."""
         super().__init__(*args)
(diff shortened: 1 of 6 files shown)


pierrot0 (approved):
Sweet, I like it!

python/mlcroissant/mlcroissant/_src/core/dataclasses.py
```diff
@@ -0,0 +1,113 @@
...
+        self.cardinality = cardinality
+        self.description = description
+        self.from_jsonld = from_jsonld
+        self.input_types = input_types
+        self.to_jsonld = to_jsonld
+        self.required = required
+        self.url = url
+
+    def call_from_jsonld(self, ctx: Context, value: Any):
+        """Calls `from_jsonld` in the field."""
```
ccl-core, line 39: If None, will return the field itself?

ccl-core (approved):
I like this a lot!

benjelloun (approved):
Looks great! 

Down the line, we could even derive some of this information from turtle definitions.

python/mlcroissant/mlcroissant/_src/core/dataclasses.py
marcenacp: If None, I return None

PR #584, opened 2024-03-03
Improve error messages with badly formatted `@id`s

ccl-core:


marcenacp (changes requested):
Thanks!

python/mlcroissant/mlcroissant/_src/core/json_ld.py
```diff
@@ -167,13 +169,31 @@ def recursively_populate_jsonld(entry_node: Json, id_to_node: dict[str, Json]) -
...
 def expand_jsonld(data: Json, ctx: Context) -> Json:
     """Expands a Croissant JSON to a nested JSON-LD with expanded.
 
     For this we use RDFLib. RDFLib expands the CURIE of the form "rdf:type" into their
     full expression, but RDFLib also flattens the JSON-LD in a list of nodes. We then
     need to reconstruct the hierarchy.
     """
+    if not check_valid_ids(data=data, ctx=ctx):
+        ctx.issues.add_error(f"There are wrong ids in this dataset. {WRONG_ID_MSG}")
+        raise ValidationError(ctx.issues.report())
```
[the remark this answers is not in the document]
ccl-core: Sure, done.

python/mlcroissant/mlcroissant/_src/tests/graphs/1.0/distribution_bad_id/metadata.json
```diff
@@ -0,0 +1,91 @@
...
+          "@id": "a-record-set/first-field",
+          "@type": "cr:Field",
+          "name": "first-field",
+          "dataType": "sc:Integer",
+          "source": {
+            "extract": {
+              "column": "a-column"
+            },
+            "fileObject": {
+              "@id": "distribution id"
```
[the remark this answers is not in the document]
ccl-core: Good point, done.

python/mlcroissant/mlcroissant/_src/core/json_ld.py
```diff
@@ -167,13 +169,31 @@ def recursively_populate_jsonld(entry_node: Json, id_to_node: dict[str, Json]) -
     return entry_node
 
 
+def check_valid_ids(data: Json, ctx: Context) -> bool:
+    """Checks that the given json contains valid `@id`s."""
+    if hasattr(data, "items"):
+        for k, v in data.items():
+            if k == "@id" and re.match(r".*\s+.*", v):
+                ctx.issues.add_error(f"The dataset contains a wrong `@id`: {v}.")
```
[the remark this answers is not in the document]
ccl-core: Done.

python/mlcroissant/mlcroissant/_src/core/json_ld.py
```diff
@@ -167,13 +169,30 @@ def recursively_populate_jsonld(entry_node: Json, id_to_node: dict[str, Json]) -
     return entry_node
 
 
+def check_valid_ids(data: Json, ctx: Context) -> bool:
+    """Checks that the given json contains valid `@id`s."""
+    if hasattr(data, "items"):
+        for k, v in data.items():
+            if k == "@id" and re.match(r".*\s+.*", v):
```
marcenacp, line 176: nit: Compiled regular expressions can run twice as fast:

```
python -c 'import time; import re; start = time.time(); [re.match(r".*\s+.*", "foo bar") for _ in range(100000)]; print(time.time() - start)'
0.3000805377960205

python -c 'import time; import re; start = time.time(); regex = re.compile(r".*\s+.*"); [regex.match("foo bar") for _ in range(100000)]; print(time.time() - start)'
0.1538679599761963
```

python/mlcroissant/mlcroissant/_src/tests/graphs/0.8/distribution_bad_type/output.txt
```diff
@@ -1,2 +1,4 @@
 Found the following 1 error(s) during the validation:
-  -  "a-csv-table" should have an attribute "@type": "https://schema.org/FileObject" or "@type": "https://schema.org/FileSet". Got https://schema.org/WRONG_TYPE_HERE instead.
\ No newline at end of file
+  -  "a-csv-table" should have an attribute "@type": "https://schema.org/FileObject" or "@type": "https://schema.org/FileSet". Got https://schema.org/WRONG_TYPE_HERE instead.
+Found the following 1 warning(s) during the validation:
+  -  `distribution` is empty for this dataset. If this is not expected, this might be because of bad `@id`s in the jsonld. Note that valid `@id`s cannot contain any whitespace characters to be successfully parsed by the rdflib.
```
marcenacp, line 4: This is 0.8 but still mentions `@id`

python/mlcroissant/mlcroissant/_src/tests/graphs/1.0/distribution_bad_type/output.txt
```diff
@@ -1,2 +1,4 @@
 Found the following 1 error(s) during the validation:
-  -  "a-csv-table" should have an attribute "@type": "http://mlcommons.org/croissant/FileObject" or "@type": "http://mlcommons.org/croissant/FileSet". Got https://schema.org/WRONG_TYPE_HERE instead.
\ No newline at end of file
+  -  "a-csv-table" should have an attribute "@type": "http://mlcommons.org/croissant/FileObject" or "@type": "http://mlcommons.org/croissant/FileSet". Got https://schema.org/WRONG_TYPE_HERE instead.
+Found the following 1 warning(s) during the validation:
+  -  `distribution` is empty for this dataset. If this is not expected, this might be because of bad `@id`s in the jsonld. Note that valid `@id`s cannot contain any whitespace characters to be successfully parsed by the rdflib.
```
marcenacp, line 4: The distribution has both a bad type and a bad `@id`?

ccl-core: No, the distribution is empty for the dataset, because it has a wrong type. The error just states that it _might_ be because of bad `@id` in the distribution (but not only, as in this case).

python/mlcroissant/mlcroissant/_src/tests/graphs/0.8/distribution_bad_type/output.txt
ccl-core: In this case, the distribution is empty because the file in the distribution has a bad type and is not registered under `distribution` (as below). The error message just states that empty distributions *can* be due to to wrong `@id`s. As one can add `@id`s for 0.8 datasets too (even if they are not mandatory as in 1.0), I added the error message for both versions. Do you think we should have it only for 1.0?

python/mlcroissant/mlcroissant/_src/core/json_ld.py
ccl-core: Sure, done.

ccl-core:
@marcenacp left a few comments. Let me know if you prefer me to avoid adding the `If this is not expected, this might be because of bad `@id`s in the jsonld. Note that valid `@id`s cannot contain any whitespace characters to be successfully parsed by the rdflib.` message when `distribution` is empty. If you think this is confusing, maybe taking it out is best.

marcenacp:
@ccl-core My fear is that it confuses the user to:

1. Mention rdflib (this is an implementation detail and the user shouldn't know about it).
2. Have the message in cases where it's not relevant.

However, I think the `if not check_valid_ids: issues.add_error` is great. Isn't it not sufficient?

python/mlcroissant/mlcroissant/_src/tests/graphs/1.0/distribution_bad_type/output.txt
goeffthomas: @marcenacp Just passing through to mention that this sort of dovetails with the conversation re: encoding. Rather than worry about what `rdflib` can/does support, it might be clearer to communicate that `@id`s are just URIs, and thus can have whitespace if URL encoded as `%20`.

ccl-core: Agreed that we do not need to cite the rdflib -- just committed a newer version of the code.

However, from a few tests I run myself, an `@id` such as `data%20frame` raises a `GenerationError`: it seems to be caused by `rdflib.graph.parse()` in [mlcroissant/_src/core/json_ld.py](https://github.com/mlcommons/croissant/blob/main/python/mlcroissant/mlcroissant/_src/core/json_ld.py) (it will expand it as if it would be a blank node identifier starting with `_` (e.g. `_:Na83c277b00434aa990d19a078d3ae785`) as it does when no `@id` is given, such in the case of `Metadata` or `Dataset`.

So URL-escaping the spaces doesn't seem to be a workaround, we actually can't have them, unless there is a fix (e.g. some parameter that we can tune in the `rdflib`, but from a quick search I couldn't find any obvious way to do so).

From what I see, however, spaces seem to be the only reserved characters  that cause problems out of all [URIs reserved characters](https://documentation.mapp.com/1.0/en/url-encoding-and-what-characters-are-valid-in-a-uri-36147771.html#id-.URLEncodingandwhatcharactersarevalidinaURIv1.0-EncodinganemailaddressinaURI).

python/mlcroissant/mlcroissant/_src/core/json_ld.py
```diff
@@ -167,13 +169,34 @@ def recursively_populate_jsonld(entry_node: Json, id_to_node: dict[str, Json]) -
...
+
 def expand_jsonld(data: Json, ctx: Context) -> Json:
     """Expands a Croissant JSON to a nested JSON-LD with expanded.
 
     For this we use RDFLib. RDFLib expands the CURIE of the form "rdf:type" into their
     full expression, but RDFLib also flattens the JSON-LD in a list of nodes. We then
     need to reconstruct the hierarchy.
     """
+    if not check_valid_ids(data=data, ctx=ctx):
+        ctx.issues.add_error("There are wrong ids in this dataset.")
```
marcenacp, line 199: This is not necessary as `check_valid_ids` already populate issues?

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/metadata.py
```diff
@@ -391,6 +391,8 @@ def from_jsonld(cls, ctx: Context, metadata: Json) -> Metadata:
                     f' "{constants.SCHEMA_ORG_FILE_SET(ctx)}". Got'
                     f" {distribution_type} instead."
                 )
+        if not distribution:
+            ctx.issues.add_warning("`distribution` is empty for this dataset.")
```
marcenacp, line 395: I would just remove this warning as the other messages are more explicit. WDYT?

marcenacp (changes requested):

python/mlcroissant/mlcroissant/_src/core/json_ld.py
```diff
@@ -167,13 +169,34 @@ def recursively_populate_jsonld(entry_node: Json, id_to_node: dict[str, Json]) -
     return entry_node
 
 
+def check_valid_ids(data: Json, ctx: Context) -> bool:
+    """Checks that the given json contains valid `@id`s."""
+    if hasattr(data, "items"):
```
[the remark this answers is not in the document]
ccl-core: Done.

python/mlcroissant/mlcroissant/_src/core/json_ld.py
```diff
@@ -167,13 +169,34 @@ def recursively_populate_jsonld(entry_node: Json, id_to_node: dict[str, Json]) -
...
+        for k, v in data.items():
+            if k == "@id" and re.match(_ID_REGEX, v):
+                ctx.issues.add_error(
+                    f"The dataset contains a wrong `@id`: '{v}'. Note that currently we"
+                    " do not support `@id`s containing whitespaces (not even"
+                    " URL-escaped)."
+                )
+                return False
```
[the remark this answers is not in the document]
ccl-core: Done.

python/mlcroissant/mlcroissant/_src/core/json_ld.py
ccl-core: Done.

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/metadata.py
ccl-core: Sure, done.

python/mlcroissant/mlcroissant/_src/tests/graphs/1.0/distribution_bad_type/output.txt
goeffthomas: I'm just a passerby, so don't let my comments hold up any progress here. I'm making a mental note to investigate this on my end. For context, at Kaggle, we rely as much as possible on what users provided to avoid baking too much "magic" into how we represent Croissant from their data. As a result, `@id` is built from the filename, which can have whitespace (among other characters). Ideally, we can just use those as-is, though I understand that if dependency issues are getting in the way, it may be unavoidable.

Bigger picture, this seems like kind of a rough thing for our users to have to be aware of about our tooling (which departs from the `@id` spec from JSON-LD itself as simply URIs which should support URL encoding). Again, not something to hold up any particular merge (including this one), but more to say that I think we may want to go after it sooner rather than later. To that point, I will be tinkering on my end with a Kaggle dataset that has whitespace in the file name (and `@id`) to see what options we may have. Hopefully there's a way but it's just not well-documented? 🙏 🤞

marcenacp (approved):
Thanks!

PR #587, opened 2024-03-06
Use `mlc_dataclasses` for all nodes (PEP 0681).

marcenacp:
image


pierrot0 (approved):
So this is not confused with the spec, could you add a link towards the spec somewhere towards the top of the generated page please

ccl-core (approved):

PR #594, opened 2024-03-12
Accept Integer/Number/Text for `sc:version`.

marcenacp:
Fixes: https://github.com/mlcommons/croissant/issues/593

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/structure_graph/nodes/metadata.py
+++ b/python/mlcroissant/mlcroissant/_src/structure_graph/nodes/metadata.py
@@ -229,7 +229,7 @@ class Metadata(Node):
     version: str | None = mlc_dataclasses.jsonld_field(
         default=None,
         description="The version of the dataset following the requirements below.",
-        input_types=[SDO.Number, SDO.Text],
+        input_types=[SDO.Integer, SDO.Number, SDO.Text],
         url=constants.SCHEMA_ORG_VERSION,
     )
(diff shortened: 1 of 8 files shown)


python/mlcroissant/mlcroissant/_src/structure_graph/nodes/metadata.py
```diff
@@ -229,7 +229,7 @@ class Metadata(Node):
     version: str | None = mlc_dataclasses.jsonld_field(
         default=None,
         description="The version of the dataset following the requirements below.",
-        input_types=[SDO.Number, SDO.Text],
+        input_types=[SDO.Integer, SDO.Number, SDO.Text],
```
ccl-core, line 232: Would it make sense to have a test to check that both `"1.0"`, `1`, `1.0.0` and `1.0` are accepted?

ccl-core (approved):
Thanks!

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/metadata.py
marcenacp: There's a [parametrized test](https://github.com/mlcommons/croissant/blob/e2e71a39ca106a99145f7faafe5de38c0b58cff3/python/mlcroissant/mlcroissant/_src/structure_graph/nodes/metadata_test.py#L39) about this :)

PR #598, opened 2024-03-12
Finish aligning mlcroissant with Croissant 1.0.

marcenacp:
I still see a few discrepancies between croissant and mlcroissant:

- RecordSet doesn’t have dataType but this is needed for sc:Enumeration.
- DataSource.format should be in DataSource.transform as it used to be? We did this change a while ago, notably because it was useful to guarantee an order between several transformations
- parentField and references have the same type but they have different types in the fact:
  - parentField => `{"source": …, "references": …}`
  - references => `{"@id": …}`
- RecordSet.examples have type JSON and URL but I’m not sure how it can be expressed in JSON-LD


docs/mlcroissant/Field.html
```diff
@@ -0,0 +1,139 @@
...
+        <a href="./Field.html">Field</a>
+    </nav>
+    <h1>Croissant &#129360</h1>
+    Croissant specifications as implemented by <code><a href="https://pypi.org/project/mlcroissant">mlcroissant</a></code>.
+    <br>
+    For the actual specifications, please refer to the <a href="https://mlcommons.org/croissant/1.0">Croissant 1.0 standard.</a>.
+</header>
+<body>
+    <h3>Field</h3>
+    <span>Nodes to describe a dataset Field.</span>
```
ccl-core, line 22: Is this a copy paste from the specs? Or are we adding more content? If the former, than maybe it makes more sense to just add a link to the specs here, so that we don't need to be sure that this docs get updated?

ccl-core: Oh, sorry -- this is generated by scripts/documentation.py no?

marcenacp: The HTML files mirror the specs **as implemented by mlcroissant**. This allows to easily compare mlcroissant to croissant as I hope it will get deployed to github.io

ccl-core (approved):
Thanks!

docs/mlcroissant/Field.html
marcenacp: > Oh, sorry -- this is generated by scripts/documentation.py no?

Yes, exactly! I hope it will get deployed to github.io. Let's try it out :)

PR #603, opened 2024-03-14
Fix for issues in Specifications and README.md

dominik-kuhn:
In this PR, a few issues in the spec files are addressed and fixed. Please review the changes and tell me, if this is the right way :)

Proposed changes:
- consistent and correct formatting in `croissant-spec.md`, `croissant-rai-spec.md` and `README.md`.
- fixed some compact IRI prefixes
- migrated all http links to https
- added some hyperlinks to expectedtypes


docs/croissant-spec.md
```diff
@@ -297,367 +274,277 @@ A "foreign key" reference on column "movie\_id" from a "ratings" table to a "mov
...
-   </td>
-   <td>The date the dataset was published.
-   </td>
+    <td>@type</td>
+    <td><a href="http://schema.org/Text">Text</a></td>
+    <td>ONE</td>
+    <td>The type of a croissant dataset must be <a href="http://schema.org/Dataset">schema.org/Dataset</a>.</td>
+  </tr>
+  <tr>
+    <td><a href="https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#conformsTo">@dct:conformsTo</a></td>
```
dominik-kuhn, line 333: Shouldnt that be 'dct:conformsTo' without an '@' instead of '@dct:conformsTo'?

benjelloun (changes requested):
Hi Dominik,

Thank you for these great fixes!

My only reservation is around changing the namespace IRIs from http to https. Historically, schema.org has used http, and now supports both. It might make sense to switch to https for Croissant as well, but we should have a separate discussion and get consensus on that before making the change in the documentation.

For now, can you please keep these IRIs as is? (I realize they are not 100% consistent in the specs, which we should fix as well.)

Thank you,
Omar

dominik-kuhn:
Ok, I reverted the http to https IRI changes and removed the @ :) should I try to fix the IRI consistency in this PR?

benjelloun (approved):
Thanks!

dominik-kuhn:
feel free to merge this PR :) I see that there are a few other Prs coming, that are addressing the same issues, this PR addresses ;)

PR #605, opened 2024-03-15
Systematically check that JSON-LD types == Python types.

marcenacp:
- Add more documentation.
- Isolate logic to cast types in dedicated functions.
- Check types (which allowed me to find some issues in the code).
- Add unit tests.


python/mlcroissant/README.md
```diff
@@ -127,6 +127,19 @@ class RecordSet(Node):
     )
 ```
 
+**Example 3**: specify a version:
```
ccl-core, line 130: NIT I would probably say something about the fact that if none given it is V1.
And that, in some cases, it is important to select another version, like if a property is only available for a specific version, like below

python/mlcroissant/mlcroissant/_src/core/versions.py
```diff
@@ -0,0 +1,34 @@
+"""Versions module."""
+
+from typing import Any
+
+from mlcroissant._src.core.issues import ErrorException
+from mlcroissant._src.core.issues import WarningException
+
+
+def cast_version(version: Any) -> str | None:
+    """Casts the given version and returns a normalized string version.
```
ccl-core, line 10: NIT: dataset version (to not confuse this with specs version?)

python/mlcroissant/mlcroissant/_src/structure_graph/base_node.py
```diff
@@ -53,13 +62,27 @@ class Node:
     ctx: Context = dataclasses.field(default_factory=Context)
     id: str = dataclasses.field(default_factory=generate_uuid)
     name: str | None = None
-    parents: list[Node] = dataclasses.field(default_factory=list)
+    parents: list["Node"] = dataclasses.field(default_factory=list)
```
ccl-core, line 65: Out of curiosity, why is this needed?

ccl-core (approved):

python/mlcroissant/README.md
marcenacp: Done.

By default, all versions are supported, not just v1.

python/mlcroissant/README.md
```diff
@@ -127,6 +127,19 @@ class RecordSet(Node):
...
+
+```python
+@mlc_dataclasses.dataclass
+class Field(Node):
+    is_enumeration: bool | None = mlc_dataclasses.jsonld_field(
+        default=None,
+        input_types=[SDO.Boolean],
+        url=constants.ML_COMMONS_IS_ENUMERATION,
+        versions=[CroissantVersion.V_0_8],   # `is_enumeration` will only be valid for v0.8, not v1.0
```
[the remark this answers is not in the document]
marcenacp: Done.

python/mlcroissant/mlcroissant/_src/core/versions.py
marcenacp: Done.

python/mlcroissant/mlcroissant/_src/structure_graph/base_node.py
marcenacp: I removed `from __future__ import annotations` in order for [`dataclasses.field`](https://docs.python.org/3/library/dataclasses.html#dataclasses.Field)'s types to be populated with types and not strings. Without this annotation, we need to use strings.

marcenacp:
Thanks for the review, @ccl-core!

python/mlcroissant/README.md
marcenacp: PS: I also added documentation in python/mlcroissant/mlcroissant/_src/core/dataclasses.py (Metadata's docstring).

PR #619, opened 2024-03-27
Add a warning in case the user deviates from the standard JSON-LD `@context`

marcenacp:
With @goeffthomas, we realized this error is common, so we look for a way to enforce it. This is a proposal.

```
$ mlcroissant validate --jsonld https://www.kaggle.com/datasets/kaggle/meta-kaggle/croissant/download

W0327 07:30:34.631297 140699768117056 rdf.py:77] WARNING: The JSON-LD `@context` is not standard. Refer to the official @context (e.g., from the example datasets in https://github.com/mlcommons/croissant/tree/main/datasets/1.0). The different keys are: {'isEnumeration', 'wd', 'dataCollection', 'dataBiases', 'personalSensitiveInformation'}
```

The problem is that the so-called "standard datasets" in the warning message also don't have a proper `@context`, so I sent PR https://github.com/mlcommons/croissant/pull/621.

```
$ mlcroissant validate --jsonld https://raw.githubusercontent.com/mlcommons/croissant/main/datasets/1.0/titanic/metadata.json

W0327 07:31:57.376785 140040957675328 rdf.py:77] WARNING: The JSON-LD `@context` is not standard. Refer to the official @context (e.g., from the example datasets in https://github.com/mlcommons/croissant/tree/main/datasets/1.0). The different keys are: {'dataCollection', 'wd', 'personalSensitiveInformation', 'dataBiases'}
W0327 07:31:57.444254 140040957675328 datasets.py:32] Found the following 1 warning(s) during the validation:
  -  [Metadata(Titanic)] Property "https://schema.org/datePublished" is recommended, but does not exist.
I0327 07:31:57.444491 140040957675328 validate.py:53] Done.
```

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/core/rdf.py
+++ b/python/mlcroissant/mlcroissant/_src/core/rdf.py
@@ -6,6 +6,7 @@ import dataclasses
 import functools
 from typing import Any
 
+from absl import logging
 from rdflib import term
 
 from mlcroissant._src.core.types import Json
@@ -71,6 +72,15 @@ class Rdf:
     def from_json(cls, ctx, json: Json) -> Rdf:
         """Creates a `Rdf` from JSON."""
         context = get_context(json)
+        # Keys that are in the standard @context, but not in the current @context.
+        different_keys = make_context(ctx).keys() - context.keys()
+        if different_keys:
+            logging.warning(
+                "WARNING: The JSON-LD `@context` is not standard. Refer to the"
+                " official @context (e.g., from the example datasets in"
+                " https://github.com/mlcommons/croissant/tree/main/datasets/1.0). The"
+                f" different keys are: {different_keys}"
+            )
         return cls(context=context)
 
     @functools.cache
--- a/python/mlcroissant/mlcroissant/_src/core/rdf_test.py
+++ b/python/mlcroissant/mlcroissant/_src/core/rdf_test.py
@@ -1,8 +1,12 @@
 """Tests for RDF."""
 
+from unittest import mock
+
+from absl import logging
 import pytest
 
 from mlcroissant._src.core.context import Context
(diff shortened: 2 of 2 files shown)


python/mlcroissant/mlcroissant/_src/core/rdf.py
```diff
@@ -71,6 +72,14 @@ class Rdf:
     def from_json(cls, ctx, json: Json) -> Rdf:
```
ccl-core, line 72: Not sure, maybe worth adding a test?

python/mlcroissant/mlcroissant/_src/core/rdf.py
```diff
@@ -71,6 +72,14 @@ class Rdf:
...
         """Creates a `Rdf` from JSON."""
         context = get_context(json)
+        difference = context.keys() - make_context(ctx).keys()
```
ccl-core, line 75: This would show only keys that are in the user defined's context and not in the standard context -- but not keys which are missing from the standard context, right?

ccl-core:
SGTM.

Maybe worth adding to the warning message that in some cases this is wanted? E.g. if the users added some custom-defined keys to their context

ccl-core (approved):
Thanks!

marcenacp:
> Maybe worth adding to the warning message that in some cases this is wanted?

Ack. I'd like to start with a simple warning.

python/mlcroissant/mlcroissant/_src/core/rdf.py
marcenacp: Done.

Good catch, I probably want to do the contrary.

python/mlcroissant/mlcroissant/_src/core/rdf.py
marcenacp: Done.

PR #621, opened 2024-03-27
Standardize the JSON-LD `@context` of all examples.

marcenacp:
I also tried to (partly) fix the migration script.


ccl-core (approved):
Thanks!

PR #628, opened 2024-04-05
fix tfds_croissant_builder.ipynb to run in colab

luisoala:
the current version of the notebook does not run through in colab

two issues

**issue 1**
install from mlcroissant from source creates following error

```bash
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-2-473943cf9267> in <cell line: 5>()
      3 
      4 from etils import epath
----> 5 import mlcroissant as mlc
      6 import requests
      7 import tensorflow_datasets as tfds

ModuleNotFoundError: No module named 'mlcroissant'
```

*hotfix:* do `!pip install mlcroissant` instead of `!pip install "git+https://github.com/${GITHUB_REPOSITORY:-mlcommons/croissant}.git@${GITHUB_HEAD_REF:-main}#subdirectory=python/mlcroissant&egg=mlcroissant[dev]"`

**issue 2**
running 

```python
%%capture --no-display
builder.download_and_prepare()
```

currently creates error

```bash
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
/usr/local/lib/python3.10/dist-packages/mlcroissant/_src/core/optional.py in _try_import(module_name, package_name)
     18     try:
---> 19         return importlib.import_module(module_name)
     20     except ImportError as exception:

21 frames
ModuleNotFoundError: No module named 'git'

The above exception was the direct cause of the following exception:

ModuleNotFoundError                       Traceback (most recent call last)
ModuleNotFoundError: Failed importing git. This likely means that the dataset requires additional dependencies that have to be manually installed (usually with `pip install GitPython`). See the optional dependencies listed in pyproject.toml.

The above exception was the direct cause of the following exception:

GenerationError                           Traceback (most recent call last)
/usr/local/lib/python3.10/dist-packages/mlcroissant/_src/operation_graph/execute.py in execute_operations_in_streaming(record_set, operations, list_of_operations, result)
    121                 result = operation(result)
    122         except Exception as e:
--> 123             raise GenerationError(
    124                 "An error occured during the streaming generation of the dataset, more"
    125                 f" specifically during the operation {operation}"

GenerationError: An error occured during the streaming generation of the dataset, more specifically during the operation Download(repo)
```
*hotfix:* add `!pip install GitPython` to pip installs at top of notebook

with the two hotfixes the full example executes

### DIFF
--- a/python/mlcroissant/recipes/tfds_croissant_builder.ipynb
+++ b/python/mlcroissant/recipes/tfds_croissant_builder.ipynb
@@ -47,7 +47,7 @@
    "source": [
     "%%capture --no-display\n",
     "# Install mlcroissant\n",
-    "!pip install mlcroissant\n",
+    "!pip install \"git+https://github.com/${GITHUB_REPOSITORY:-mlcommons/croissant}.git@${GITHUB_HEAD_REF:-main}#subdirectory=python/mlcroissant&egg=mlcroissant[dev]\" || pip install mlcroissant\n",
     "!pip install array_record\n",
     "!pip install tfds-nightly\n",
     "!pip install tensorflow\n",
--- a/python/mlcroissant/recipes/tfds_croissant_builder.ipynb
+++ b/python/mlcroissant/recipes/tfds_croissant_builder.ipynb
@@ -46,7 +46,7 @@
    "outputs": [],
    "source": [
     "%%capture --no-display\n",
-    "# Install mlcroissant from the source\n",
+    "# Install mlcroissant\n",
     "!pip install mlcroissant\n",
     "!pip install array_record\n",
     "!pip install tfds-nightly\n",
--- a/python/mlcroissant/recipes/tfds_croissant_builder.ipynb
+++ b/python/mlcroissant/recipes/tfds_croissant_builder.ipynb
@@ -47,11 +47,12 @@
    "source": [
     "%%capture --no-display\n",
     "# Install mlcroissant from the source\n",
-    "!pip install \"git+https://github.com/${GITHUB_REPOSITORY:-mlcommons/croissant}.git@${GITHUB_HEAD_REF:-main}#subdirectory=python/mlcroissant&egg=mlcroissant[dev]\"\n",
+    "!pip install mlcroissant\n",
     "!pip install array_record\n",
     "!pip install tfds-nightly\n",
     "!pip install tensorflow\n",
     "!pip install torch\n",
+    "!pip install GitPython\n",
     "!apt-get install tree"
    ]
   },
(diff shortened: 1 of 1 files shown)


python/mlcroissant/recipes/tfds_croissant_builder.ipynb
```diff
@@ -46,12 +46,13 @@
    "outputs": [],
    "source": [
     "%%capture --no-display\n",
-    "# Install mlcroissant from the source\n",
-    "!pip install \"git+https://github.com/${GITHUB_REPOSITORY:-mlcommons/croissant}.git@${GITHUB_HEAD_REF:-main}#subdirectory=python/mlcroissant&egg=mlcroissant[dev]\"\n",
+    "# Install mlcroissant\n",
+    "!pip install mlcroissant\n",
```
marcenacp, line 50: This breaks the feature that the notebook is run in the GitHub Actions on the current branch to test it end-to-end. What doesn't work with the current approach?

marcenacp: Mmmh looks like the execution times out. Probably GitHub has some issues right now?

marcenacp: It also times out in the said GitHub Actions: https://github.com/mlcommons/croissant/actions/runs/8556498478/job/23446447330 (currently broken on main)

marcenacp: The command works locally so that sounds more like an issue with Colab

luisoala: if its a temporary timeout issue we can ignore i think

it also affects this notebook https://colab.research.google.com/github/mlcommons/croissant/blob/main/python/mlcroissant/recipes/introduction.ipynb which is linked directly to colab here https://github.com/mlcommons/croissant/tree/main/python/mlcroissant

Screenshot from 2024-04-05 12-19-07

if we expect users to inspect the recipes in browser through colab, another fix would be to use `||` in bash to catch install with `pip install mlcroissant` in case these timeouts occur

`!pip install "git+https://github.com/${GITHUB_REPOSITORY:-mlcommons/croissant}.git@${GITHUB_HEAD_REF:-main}#subdirectory=python/mlcroissant&egg=mlcroissant[dev]" || pip install mlcroissant`

marcenacp: OK, so `%%capture --no-display` was obfuscating the actual errors. I'm trying a fix that would work for both humans and the CI: https://github.com/mlcommons/croissant/pull/630

luisoala: works great ! closing this for #630

PR #642, opened 2024-05-02
update readme code example to new hf and croissant api

luisoala:
- update hf link to fmnist
- add record_set_ids and file_format args
- add builder.download_and_prepare() else the train, test example does not work


luisoala:
@mkuchnik see these updates to minimal example

marcenacp (approved):
Thanks!!

PR #652, opened 2024-05-15
Remove flores notebook from the automatically checked notebook.

marcenacp:
Reason: it seems that torchdata is not maintained anymore ([source](https://github.com/pytorch/data?tab=readme-ov-file#torchdata-see-note-below-on-current-status)) and it fails with torch versions in [2.1, 2.2, 2.3].

@mkuchnik FYI, we excluded the notebook from the GitHub Actions because it has been failing in the last >2 weeks. We unsuccessfully tried to fix the versions, probably you'll know better. Thanks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,16 +80,7 @@ jobs:
           do
             if [ "$notebook" = "recipes/flores200_datapipes.ipynb" ]
             then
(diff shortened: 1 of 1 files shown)


ccl-core (approved):

marcenacp:
(Failing non hermetic tests due to OpenML API)

PR #658, opened 2024-05-17
Use regex-based version casting that accept 

ccl-core:
Remove warning when version doesn't conform to MAJOR.MINOR.PATCH

Ref. #https://github.com/mlcommons/croissant/issues/609


python/mlcroissant/mlcroissant/_src/core/issues.py
```diff
@@ -34,6 +34,7 @@ class Issues:
     We use sets to represent errors and warnings to avoid repeated strings.
     """
 
+    ignore_warnings: bool = False
```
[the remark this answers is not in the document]
marcenacp: +1 to put it in `report`.

python/mlcroissant/mlcroissant/_src/core/issues_test.py
```diff
@@ -42,3 +42,22 @@ def test_issues():
...
+    assert not issues.errors
+    assert not issues.warnings
+
+    metadata = Metadata(id="abc", name="abc")
+    file_object = FileObject(id="xyz", name="xyz")
+    file_object.parents = [metadata]
+    issues.add_error("foo", metadata)
+    issues.add_warning("bar", file_object)
+
+    print(issues.report())
```
marcenacp, line 58: Remove?

ccl-core:
That's right, this would actually address the "However, adding a generic --ignore framework is probably a good first step, so I'm opening this issue." part of the issue, but I think a more long-term solution might be needed to silence only a subset of the warnings.

python/mlcroissant/mlcroissant/_src/core/versions.py
```diff
@@ -19,10 +19,9 @@ def cast_version(version: Any) -> str | None:
     if isinstance(version, str):
         numbers = version.split(".")
         are_not_all_numbers = any(not number.isnumeric() for number in numbers)
-        if len(numbers) != 3 or are_not_all_numbers:
+        if are_not_all_numbers:
             raise WarningException(
-                f"Version doesn't follow MAJOR.MINOR.PATCH: {version}. For more"
-                " information refer to: https://semver.org/spec/v2.0.0.html"
+                f"Version contains non-numeric characters: {version}."
```
ccl-core, line 24: Keeping the warning for non-numeric versioning. Reasonable? Or shall I remove this as well?

ccl-core:
@marcenacp updated the PR -- PTAL, thanks! :)

python/mlcroissant/mlcroissant/_src/core/versions.py
marcenacp: We could fully implement the [semantic versioning standard](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions).

[`semver`](https://pypi.org/project/semver) is a python library that does exactly this. However it seems we only need to check the compatibility with the standard, so we could also just reuse their [regular expression](https://github.com/python-semver/python-semver/blob/c47774db88386f7db79e5ec9fe3672c07ed7c063/src/semver/version.py#L81-L104).

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/metadata_test.py
```diff
@@ -101,17 +101,9 @@ def test_invalid_version(version, expected_error):
 @pytest.mark.parametrize(
     ["version", "expected_warning"],
     [
-        [
```
marcenacp, line 104: Still keep this?

marcenacp: (Those should fail)

marcenacp: Above, you could add that the versions `1` and `1.2` now succeed

marcenacp (approved):

python/mlcroissant/mlcroissant/_src/core/versions.py
marcenacp: Also, I think that it's worth mentioning the semantic versioning standard: https://semver.org/spec/v2.0.0.html, and that minor and patch can be omitted.

marcenacp (changes requested):
See comments.

marcenacp (commented):
Thanks!

marcenacp (approved):

PR #665, opened 2024-05-30
update readme with paper proceedings info

luisoala:
acm proceedings are published now here https://dl.acm.org/doi/proceedings/10.1145/3650203


benjelloun (approved):
Thanks!

PR #670, opened 2024-06-04
Add four record sets to anthropic hh hlhf

ccl-core:


python/mlcroissant/mlcroissant/_src/operation_graph/graph.py
```diff
@@ -115,7 +115,7 @@ def _add_operations_for_git(
     for successor in node.successors:
         if isinstance(successor, FileSet):
             fields = tuple(
-                field for field in node.recursive_successors if isinstance(field, Field)
+                field for field in successor.recursive_successors if isinstance(field, Field)
```
ccl-core, line 118: Previously, we were adding all fields originating from `repo`, which was not correct and created problem down the line. In the case of this dataset, self.fields of the Read operation that should open and read `harmless-base-files` included also a lot of non-relevant fields from `red-team-attempts`. This was generating a ValueError `The fields use several reading methods`, as some of these fields use JSON and some others use JSON_LINES are reading method.

https://github.com/mlcommons/croissant/blob/7000545c1352979f7694a786aadd67cdc74f0269/python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py#L57)

ccl-core:
Removed splits record set (to be added in the next PR).

marcenacp (approved):

PR #680, opened 2024-06-05
Fix small bugs for splits.

ccl-core:
* Add `wd:` to the context when needed
* Fix small bug in regex transformation when dealing with paths

### DIFF
--- a/datasets/1.0/audio_test/metadata.json
+++ b/datasets/1.0/audio_test/metadata.json
@@ -42,7 +42,8 @@
     "separator": "cr:separator",
     "source": "cr:source",
     "subField": "cr:subField",
-    "transform": "cr:transform"
+    "transform": "cr:transform",
+    "wd": "https://www.wikidata.org/wiki/"
   },
   "@type": "sc:Dataset",
   "name": "audio_test",
--- a/datasets/1.0/audio_test/metadata.json
+++ b/datasets/1.0/audio_test/metadata.json
@@ -42,8 +42,7 @@
     "separator": "cr:separator",
     "source": "cr:source",
     "subField": "cr:subField",
-    "transform": "cr:transform",
-    "wd": "https://www.wikidata.org/wiki/"
+    "transform": "cr:transform"
   },
   "@type": "sc:Dataset",
   "name": "audio_test",
--- a/datasets/1.0/huggingface-anthropic-hh-rlhf/metadata.json
+++ b/datasets/1.0/huggingface-anthropic-hh-rlhf/metadata.json
@@ -323,7 +323,7 @@
               "fileProperty": "fullpath"
             },
             "transform": {
-              "regex": ".*(train|test)\\.jsonl\\.gz$"
+              "regex": ".*/(train|test)\\.jsonl\\.gz$"
             }
           }
         },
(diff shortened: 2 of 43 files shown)


datasets/1.0/audio_test/metadata.json
```diff
@@ -42,7 +42,8 @@
     "separator": "cr:separator",
     "source": "cr:source",
     "subField": "cr:subField",
-    "transform": "cr:transform"
+    "transform": "cr:transform",
+    "wd": "https://www.wikidata.org/wiki/"
```
marcenacp, line 46: wd shouldn't be in the default context. I'm of the opinion that it should be added only when needed (the context is long enough). We can discuss this.

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
```diff
@@ -34,6 +35,8 @@ def _apply_transform_fn(value: Any, transform: Transform, field: Field) -> Any:
     """Applies one transform to `value`."""
     if transform.regex is not None:
         source_regex = re.compile(transform.regex)
+        if isinstance(value, pathlib.PurePath):
```
marcenacp, line 38: Always do:

```python
value = os.fspath(value)
```

?

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field_test.py
```diff
@@ -1,5 +1,6 @@
 """field_test module."""
 
+import pathlib
```
marcenacp, line 3: Use epath.Path everywhere?

datasets/1.0/huggingface-anthropic-hh-rlhf/metadata.json
```diff
@@ -266,6 +303,31 @@
       "@id": "harmless-base",
       "name": "harmless-base",
       "field": [
+        {
+          "@type": "cr:Field",
+          "@id": "harmless-base/split",
```
marcenacp, line 308: Are the fields repeated between the RecordSets?

marcenacp (changes requested):
Great, thanks!

datasets/1.0/huggingface-anthropic-hh-rlhf/metadata.json
ccl-core: Exactly -- is there a better way to express this with less redundancy?

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
ccl-core: Done.

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field_test.py
ccl-core: In tests sure. In `field.py` not sure, I'd need to convert `fullpath` to be epath.Path as it is a `pathlib.PurePath` (https://github.com/mlcommons/croissant/blob/0f95e04763557929e4f4c6711c108c0d9cf7b818/python/mlcroissant/mlcroissant/_src/core/path.py#L24)

marcenacp: Agreed.

datasets/1.0/huggingface-anthropic-hh-rlhf/metadata.json
marcenacp: We could parametrize the config name. But that'd be a feature in itself, so no need to do this in this PR. Good to know we have a dataset to develop this feature!

marcenacp (approved):
Thanks!

datasets/1.0/audio_test/metadata.json
ccl-core: Sure. Done.

PR #699, opened 2024-06-13
Fix regex transform failure on FileProperty.fullpath

brendon-boldt:
Also add a test case which fails without the fix.  I don't know if this is the best place to cast `PurePath` to `str`, but it has the least impact here while still solving the problem.


brendon-boldt:
recheck

ccl-core:
Hi @brendon-boldt , and thank you for this PR! The bug has been fixed in #680 :)

brendon-boldt:
Whoops.  I didn't do my due diligence in checking what was already open.

PR #704, opened 2024-06-24
Fix discrepancy in rdflib output with node.data

ccl-core:


ccl-core:
Close as we solved the issue otherwise.

PR #705, opened 2024-06-25
Add URLs to pyproject.toml

PGijsbers:
This is useful since it shows up on the PyPI sidebar for the package, which means that users will be able to easily navigate to the github, bugs, and documentation in a standardized way. Not sure what the most appropriate documentation pages would be at this point.


pierrot0 (approved):
Thanks Pieter!

PGijsbers:
Sorry, it seems I do not have write access. Could you merge this then?:) thanks

ccl-core:
Sure!

PR #713, opened 2024-07-16
When creating an `mlc.Metadata` object, share the graph with all nodes.

marcenacp:


marcenacp:
Red tests are unrelated to this PR.

PR #714, opened 2024-07-18
DataTypes should be URIRef

ccl-core:
We should always convert DataTypes from string format to URIRef.

Currently, we do it only in some cases, it should be done consistently.


marcenacp (approved):

ccl-core:
Failing tests unrelated to this PR

PR #716, opened 2024-07-18
Implement filtering in the case of filename regular expression and add a test for this feature.

marcenacp:
In this PR:

- I introduce in mlcroissant API a light version of filtering by allowing a dictionary that maps a field's ID to the value we want to filter.
- I add a new test dataset that contains 2 parquet files (`train.parquet` for train, `test.parquet` for test). The regular expression is still `*.parquet`, so we test that passing the filter `{'data/split': 'train'}` still only reads elements from `train.parquet`.
- For now, the coverage is very limited (basically only the cases we could encounter in Hugging Face datasets), so I just throw a bunch of `NotImplementedErrors`.


ccl-core:
That's great, thanks Pierre!

NIT, as IIUC currently it works only for single values (filters can't be `{'split': ['test1', 'test2']}`) maybe it makes sense to add some checks to ensure only single values are passed? So that it doesn't fail at `_regex_from_value` when you `re.escape(value)`

ccl-core (approved):
Thanks!

marcenacp:
> NIT, as IIUC currently it works only for single values (filters can't be `{'split': ['test1', 'test2']}`) maybe it makes sense to add some checks to ensure only single values are passed? So that it doesn't fail at `_regex_from_value` when you `re.escape(value)`

Done.

PR #717, opened 2024-07-19
Fix broken Unit tests.

ccl-core:
Fix errors in dataset_test.py:

* Remove `TGSR_test` from `test_nonhermetic_loading_1_0`, it is failing because underlying data changed.
* Move `audio_test` to `test_hermetic_loading` as it doesn't contain data from the internet.
* Added `soxr==0.4.0b1` dependency to `pyproject.toml` to solve an issue with [librosa](https://github.com/librosa/librosa/issues/1831)

### DIFF
--- a/python/mlcroissant/pyproject.toml
+++ b/python/mlcroissant/pyproject.toml
@@ -49,6 +49,7 @@ dev = [
   "pylint",
   "pytest",
   "pytype",
+  "soxr==0.4.0b1",
   "torchdata",
 ]
 audio = ["librosa"]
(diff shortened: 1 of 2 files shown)


python/mlcroissant/pyproject.toml
```diff
@@ -51,7 +51,10 @@ dev = [
   "pytype",
   "torchdata",
 ]
-audio = ["librosa"]
+audio = [
+  "librosa",
+  "soxr==0.4.0b1",
```
marcenacp, line 56: Can you add a comment with the fix: https://github.com/librosa/librosa/issues/1831#issuecomment-2176274560

ccl-core: Sure, done.

PR #719, opened 2024-07-19
Apply filters to a Hugging Face dataset to avoid repeating all variants.

marcenacp:
The only issue for now is that `regex` is a regex while `includes` is a glob... So I use heuristics to convert from one to another. I think it's not a problem for hugging face datasets as we control the form they have. But it can be challenging to have a generic conversion. The best would be to use either regular expressions or glob patterns everywhere.


python/mlcroissant/mlcroissant/_src/datasets_test.py
```diff
@@ -192,13 +191,16 @@ def test_nonhermetic_loading(version, dataset_name, record_set_name, num_records
 # Non-hermetic test cases for croissant >=1.0 only (data from the internet).
 @pytest.mark.nonhermetic
 @pytest.mark.parametrize(
-    ["dataset_name", "record_set_name", "num_records"],
+    ["dataset_name", "record_set_name", "num_records", "filters"],
```
ccl-core, line 194: So here we could also e.g. have 

`["huggingface-anthropic-hh-rlhf/metadata.json", "harmless-base", 10, "train"]`

right?

ccl-core (approved):
That's cool! Thanks :)

python/mlcroissant/mlcroissant/_src/datasets_test.py
marcenacp: I haven't tried this dataset yet!

marcenacp: But potentially yes.

PR #726, opened 2024-08-30
Add splits to the huggingface-mnist dataset

ccl-core:


python/mlcroissant/mlcroissant/scripts/load_test.py
```diff
@@ -7,18 +7,18 @@
 from mlcroissant.scripts import load as load_lib
 
 
-@parametrize_version()
-def test_should_raise_when_no_record_set(version):
+def test_should_raise_when_no_record_set():
```
ccl-core, line 10: We could keep the version fixture if we remove the match of the ValueError message. WDYT?

marcenacp: I'm not sure we have to support 0.8 anymore, so I'm ok with this :)

marcenacp (approved):
Thanks!

PR #730, opened 2024-09-03
Allow to parallelize operations in mlcroissant with Apache Beam.

marcenacp:


python/mlcroissant/mlcroissant/_src/datasets.py
```diff
@@ -163,6 +168,35 @@ def __iter__(self):
                 record_set=self.record_set, operations=operations
             )
 
+    def beam_reader(self, pipeline: beam.Pipeline):
+        """Returns an Apache Beam sink to generate the dataset using Flume/Spark/etc.
```
ccl-core, line 172: Is Flume open sourced?

python/mlcroissant/mlcroissant/_src/datasets_test.py
```diff
@@ -114,6 +117,46 @@ def load_records_and_test_equality(
     assert len(expected_records) == length
 
 
+def _equal_to_set(expected):
+    """Tests that 2 PCollections are equal as sets."""
+
```
ccl-core, line 122: I guess this is not a test, maybe "checks whether 2 PCollections ..." ?

python/mlcroissant/mlcroissant/_src/operation_graph/execute.py
```diff
@@ -124,3 +132,37 @@ def read_all_files():
...
             ) from e
+
+
+def execute_operations_in_beam(
+    pipeline: beam.Pipeline, record_set: str, operations: Operations
+):
+    """Executes operations in Beam.
+
+    This is only possible for streamable datasets, i.e. datasets that result from
+    executing a linear sequence of operations. If there are branches we'd need a more
```
ccl-core, line 143: Does this include e.g. joins too?

python/mlcroissant/mlcroissant/_src/datasets.py
```diff
@@ -163,6 +168,35 @@ def __iter__(self):
...
+            jsonld="https://huggingface.co/api/datasets/ylecun/mnist/croissant",
+        )
+        pipeline_options = PipelineOptions()
+        with beam.Pipeline(options=pipeline_options) as pipeline:
+            _ = dataset.records("mnist").beam_reader(pipeline)
+        ```
+        """
+        operations = self._filter_interesting_operations(self.filters)
+        execute_downloads(operations)
+        can_stream_dataset = all(d == 1 or d == 2 for _, d in operations.degree())
```
ccl-core, line 191: NIT: I know you explain this below, but I would mention also in the docstring of this function 1) that it raises a ValueError for non-streamable datasets, and 2) what streamable datasets are.

python/mlcroissant/mlcroissant/_src/operation_graph/execute.py
ccl-core: This would be important to know because if yes, IIUC then we would have a problem with datasets with splits?

python/mlcroissant/mlcroissant/_src/operation_graph/execute.py
```diff
@@ -124,3 +132,37 @@ def read_all_files():
...
                 f" specifically during the operation {operation}"
             ) from e
+
+
+def execute_operations_in_beam(
+    pipeline: beam.Pipeline, record_set: str, operations: Operations
+):
+    """Executes operations in Beam.
+
+    This is only possible for streamable datasets, i.e. datasets that result from
```
ccl-core, line 142: NIT: "This is only possible for streamable datasets" --> Not sure if this is clear from the code... the check for non-streamable datasets (and the consequent error) happens in beam_reader no? So what would happen here if a non-streamable dataset would be given as input here? Where would the function fail?

ccl-core (approved):
That's great work! Thanks a lot

python/mlcroissant/mlcroissant/_src/datasets.py
marcenacp: Done.

python/mlcroissant/mlcroissant/_src/operation_graph/execute.py
marcenacp: Done.

python/mlcroissant/mlcroissant/_src/datasets_test.py
marcenacp: Done.

python/mlcroissant/mlcroissant/_src/datasets.py
marcenacp: Done.

PR #731, opened 2024-09-04
More features around Beam.

marcenacp:
- Simplify pickling/hashing for nodes.
- Better public API for ReadFromCroissant.
- Add indices to each example for TFDS to use.


pierrot0 (approved):

python/mlcroissant/mlcroissant/_src/beam.py
```diff
@@ -0,0 +1,70 @@
...
+            pipeline=pipeline,
+            jsonld=jsonld,
+            record_set="default",
+        )
+    ```
+
+    Only streamable datasets can be used with Beam. A streamable dataset is a dataset
+    that can be generated by a linear sequence of operations - without joins for
+    example. This is the case for Hugging Face datasets. If there are branches, we'd
+    need a more complex Beam pipeline.
```
ccl-core, line 47: So IIUC we will change this in future work to work with non-streamable datasets (e.g. future versions of HF croissants) -- I feel if this is right, this would deserve a mention here, or a link to an issue?

python/mlcroissant/mlcroissant/_src/operation_graph/execute.py
```diff
@@ -149,12 +154,69 @@ def execute_operations_in_beam(
...
+) -> Generator[ElementWithIndex, None, None]:
+    """Computes the global index given the shard size."""
+    shard_index, element = element_with_index
+    for index_in_shard, result in enumerate(operation(element)):
+        if index_in_shard >= max_shard_size:
+            raise ValueError(
+                "WARNING: This was very unlikely, but it seems we just hit this limit"
+                " in the code. Find another way to optimize execute_operations_in_beam."
+                " Please, open a PR on GitHub to make the maintainers aware of this"
+                " issue. An actual easy fix is to compute the actual shard_sizes rather"
```
ccl-core, line 210: rm first "actual"?

python/mlcroissant/mlcroissant/_src/structure_graph/base_node.py
```diff
@@ -304,8 +304,8 @@ def there_exists_at_least_one_property(self, possible_properties: list[str]):
         return False
 
     def __hash__(self):
-        """Hashes all immutable arguments."""
-        return hash(self.uuid)
+        """Re-uses parent's hash function."""
```
ccl-core, line 307: Not sure why we would use the parent's hash?

python/mlcroissant/mlcroissant/_src/operation_graph/execute.py
```diff
@@ -149,12 +154,69 @@ def execute_operations_in_beam(
...
+    # solution to this problem would be to compute the shard_sizes in parallel of
+    # generating the records.
+    # TODO(marcenacp): compute shard_sizes explicitly instead of relying on
+    # max_shard_size.
+    max_shard_size = sys.maxsize // num_shards
     while queue_of_operations:
         operation = queue_of_operations.popleft()
         if isinstance(operation, ReadFields):
-            beam_operation = beam.ParDo(operation)
+            beam_operation = beam.ParDo(
```
ccl-core, line 182: We are 100% sure that only ReadFields operations can be leaves?

python/mlcroissant/mlcroissant/_src/operation_graph/execute.py
```diff
@@ -149,12 +154,69 @@ def execute_operations_in_beam(
...
+        raise ValueError("Could not filter files.")
+    files = list(files)  # even for large datasets, this can be handled in RAM.
+
+    # We first shard by file and assign a shard_index.
+    pipeline = pipeline | "Shard by files with index" >> beam.Create(enumerate(files))
+    num_shards = len(files)
+
+    # We don't know in advance the number of records per shards. So we just allocate the
+    # maximum number which is `sys.maxsize // num_shards`. For a large evenly
+    # distributed dataset, we have:
```
ccl-core, line 167: NIT: Maybe:

"Taking the practical case of large evenly distributed datasets on HuggingFace, we can compute the following:"

or something similar? Currently it is a bit confusing for a user reading about very specific HF numbers in the context of a generic function.

ccl-core (approved):
Thanks!!!

python/mlcroissant/mlcroissant/_src/beam.py
marcenacp: Done.

python/mlcroissant/mlcroissant/_src/operation_graph/execute.py
marcenacp: Done.

python/mlcroissant/mlcroissant/_src/operation_graph/execute.py
marcenacp: Done.

PR #736, opened 2024-09-10
Isolate a `.call()` method in operations.

marcenacp:
- [Simplify graph printing.](https://github.com/mlcommons/croissant/pull/736/commits/a3d42557f65add58fb4e2888d1aebe591451a201)
- [Isolate a .call() method in operations.](https://github.com/mlcommons/croissant/pull/736/commits/36e1d21dc536ceb150aeb6316a5a43aab13d74f8)

Later, the goal is to have the `.call()` method to populate an internal cache.

Pseudo-code for next PRs:

```python
class Operation:
  _output: Any = dataclass(init=False)

  def __call__(...):
    if self._output:
      return self._output
    result = self.call(...)
    self._output = result
    return result
```

This would allow to retrieve the result of any operation by calling `operation()` - without argument. This should hopefully simplify the logic of making data flow in the graph of operations.

In the context of the Beam pipeline, this would allow us to generate intermediary steps (like intermediary RecordSets for joins for example) more easily.

When we don't want to use the cache (e.g., in the case of streamable datasets), we can just use operation.call() as we would normally do.

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/core/graphs/utils.py
+++ b/python/mlcroissant/mlcroissant/_src/core/graphs/utils.py
@@ -5,22 +5,14 @@ import time
 import networkx as nx
 
 
-def pretty_print_graph(graph: nx.Graph, simplify=False):
+def pretty_print_graph(graph: nx.Graph):
     """Pretty prints a NetworkX graph.
 
     Warning: this function is for debugging purposes only.
 
     Args:
         graph: Any NetworkX graph.
-        simplify: Whether to print a simplified version of nodes.
     """
-    if simplify:
-        simple_graph = nx.Graph()
-        for x, y in graph.edges():
-            x = getattr(x, "uid", x)
-            y = getattr(y, "uid", y)
-            simple_graph.add_edge(x, y)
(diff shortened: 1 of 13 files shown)


python/mlcroissant/mlcroissant/_src/core/graphs/utils.py
```diff
@@ -5,22 +5,14 @@
 import networkx as nx
 
 
-def pretty_print_graph(graph: nx.Graph, simplify=False):
+def pretty_print_graph(graph: nx.Graph):
```
ccl-core, line 8: Are we sure this is not handy in some circumstances? You would obtain the same if you would remove the `simplify=True` args in the function calls below, and keep the extra simplify option here, no?

ccl-core (approved):
Thanks!

python/mlcroissant/mlcroissant/_src/core/graphs/utils.py
marcenacp: It seems I haven't found a use case for it. Worst case it's still in the git history :)

PR #741, opened 2024-09-22
[WIP] Cache the result of each operation.

marcenacp:
This will aim at more easily generating datasets in Beam - even when they contain joins.


python/mlcroissant/mlcroissant/_src/operation_graph/base_operation.py
```diff
@@ -64,12 +76,20 @@ class Operation(abc.ABC):
 
     operations: Operations
     node: Node
+    _output: OutputT = dataclasses.field(init=False, hash=False, compare=False)
```
ccl-core, line 79: Can you add _output to the docstring? :)

ccl-core (approved):
Cool! Thanks a lot :)

python/mlcroissant/mlcroissant/_src/operation_graph/base_operation.py
marcenacp: Done.

PR #742, opened 2024-09-23
Fix discrepancies with the specs

ccl-core:
1) ids and names are the same for Fields and RecordSets (see migration 202409231500.py): updated metadata and output;
2) In the `get_column` method of `Source`, we return the node's `uuid` if no extract method is specified;
3) for RecordSet specifying `data`, we look at `field.id` and not `field.name` to get the expected keys;
4) also added a filters flag to load.py.


datasets/1.0/simple-parquet/output/persons.jsonl
```diff
@@ -1,10 +1,10 @@
...
-{"name": "person1", "age": 1}
-{"name": "person2", "age": 2}
-{"name": "person3", "age": 3}
-{"name": "person4", "age": 4}
-{"name": "person5", "age": 5}
-{"name": "person6", "age": 6}
-{"name": "person7", "age": 7}
-{"name": "person8", "age": 8}
-{"name": "person9", "age": 9}
+{"persons/name": "person0", "persons/age": 0}
```
marcenacp, line 1: I think it was feature to not have the RecordSet ID here - otherwise, it's repeated:

- The user is asking for the `persons` RecordSet
- => All fields will start with `persons`

Maybe, it's a consequence of my other comment where you set `name == @id` for all datasets. It's the case for Hugging Face datasets, but it may also not be the case per the specs.

So it could be good to keep at least one dataset with `name != @id` for testing purposes. What do you think?

marcenacp (approved):
Thanks!

python/mlcroissant/mlcroissant/scripts/migrations/previous/202409231500.py
```diff
@@ -0,0 +1,18 @@
...
+the specs.
+"""
+
+
+def up(json_ld):
+    """Up function."""
+    json_copy = json_ld.copy()
+    for _, record_set in enumerate(json_copy.get("recordSet", [])):
+        if record_set["@id"] != record_set["name"]:
+            record_set["name"] = record_set["@id"]
```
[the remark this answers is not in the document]
ccl-core: You are right, it is not a requirement... but it seems to be the recommended way though, given that it is so for all given examples :)

datasets/1.0/simple-parquet/output/persons.jsonl
ccl-core: Sure! I updated all datasets because of the migration script, but good point to keep one different.

PR #744, opened 2024-09-24
Use ids to reference a field or a node.

ccl-core:
Names are optional (so they can also not be specified). We should use ids to reference nodes.


marcenacp (approved):
Great fixes. Thanks!!!

PR #747, opened 2024-09-25
Check that the mapping is valid after setting it.

marcenacp:


ccl-core (approved):
LGTM, thanks!

python/mlcroissant/mlcroissant/_src/datasets_test.py
```diff
@@ -330,3 +331,31 @@ def test_validate_filters(filters, raises):
...
+def test_check_mapping_when_the_key_does_not_exist(version):
+    dataset_name = "simple-parquet/metadata.json"
+    record_set_name = "persons"
+    with pytest.raises(ValueError, match="doesn't exist in the JSON-LD"):
+        load_records_and_test_equality(
+            version,
+            dataset_name,
+            record_set_name,
+            -1,
+            mapping={"this_UUID_does_not_exist": "/this/path/does/not/exist"},
```
ccl-core, line 346: NIT -- it would be nice to have an existing path here somehow

marcenacp: Done.

PR #757, opened 2024-10-22
Make field more robust with None/nan repeated input

ccl-core:


python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
```diff
@@ -145,6 +148,30 @@ def _extract_value(df: pd.DataFrame, field: Field) -> pd.DataFrame:
...
+        raise ValueError(
+            "Nested subfields can only be populated when the parent field exists!"
+        )
+    parent_id = field.parent.id
+    if parent_id not in result:
+        result[parent_id] = (
+            [{field.id: v} for v in value] if not _is_na(value) else [{field.id: value}]
+        )
+    else:
+        if not _is_na(value) and len(value) != len(result[parent_id]):
```
marcenacp, line 165: Could this function be simplified? Something like:

```python
value = [{field.id: None}] if _is_na(value) else value
existing_value = result.get(parent_id, value)
if not _is_na(value) and len(value) != len(existing_value):
  raise ValueError(...)
for i in range(existing_value):
  result[parent_id][i][field.id] = None if _is_na(value) else value[i]
```

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
```diff
@@ -193,9 +220,11 @@ def _get_result(row):
                 )
                 value = apply_transforms_fn(value, field=field, repeated=is_repeated)
                 if is_repeated:
-                    value = [
-                        _cast_value(self.node.ctx, v, field.data_type) for v in value
-                    ]
+                    value = (
+                        [_cast_value(self.node.ctx, v, field.data_type) for v in value]
+                        if not _is_na(value)
+                        else value
```
marcenacp, line 226: Would this function benefit an early return as well?

marcenacp (approved):
Approve to unblock but left a few comments.

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
ccl-core: Not sure I understand (we return result, not value) but I changed it to a if..elif..else, it should be more readable now :)

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
ccl-core: I simplified it a bit, hope it goes in the direction you had in mind!

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
```diff
@@ -57,7 +57,7 @@ def _apply_transform_fn(value: Any, transform: Transform, field: Field) -> Any:
             return pd.Timestamp(value).strftime(transform.format)
         else:
             raise ValueError(f"`format` only applies to dates. Got {field.data_type}")
-    elif transform.separator is not None:
+    elif transform.separator is not None and not _is_na(value):
```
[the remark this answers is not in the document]
ccl-core: Yes, done.

PR #764, opened 2024-11-12
Fix bug with jsonpath_rw and numpy arrays

ccl-core:
Loading lists from parquet using pandas deserializes them as numpy's ndarrays. However, we need lists as input to jsonpath_rw's find function (ndarrays return `IndexError` so we don't catch the exception in https://github.com/kennknowles/python-jsonpath-rw/blob/6f5647bb3ad2395c20f0191fef07a1df51c9fed8/jsonpath_rw/jsonpath.py#L436C1-L436C54)


marcenacp (approved):
Thanks!

PR #767, opened 2024-11-25
Include a prefix to the beam pipeline's stages

ccl-core:


marcenacp (approved):

ccl-core:
Updated as discussed offline, thanks!

PR #774, opened 2024-11-29
Handle non-capturing groups in regex transforms (`partial-train/*.parquet`).

marcenacp:
The initial issue handled in this PR is to cover for Hugging Face's regex pattern: `"default/(?:partial-)?(train|test)/.+parquet$"`.

Indeed, in the parquet branch, when the dataset is not entirely moved to parquet, the folder may be called `partial-`, so we have to handle those files.

The difficulty is that `transform.regex` is a regular expression while `includes` is a glob pattern, so we have to convert from one to the other.

Before this PR, the following command fails:

```bash
mlcroissant load --jsonld https://huggingface.co/api/datasets/mlfoundations/dclm-baseline-1.0-parquet/croissant --record_set default --num_records 1 --debug --filters '{"default/split": "train"}'
```

After this PR, it does succeed, although there's another regression: we download all parquet files before doing the join instead of joining file by file.

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/core/regex.py
+++ b/python/mlcroissant/mlcroissant/_src/core/regex.py
@@ -1,0 +1,88 @@
+from collections.abc import Iterable
+import itertools
+import re
+from typing import Any
+
+
+def regex_to_glob(regex: str | list[str]) -> list[str]:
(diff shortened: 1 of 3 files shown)


ccl-core (approved):
Very nice! Thank you :)

PR #775, opened 2024-11-29
Drop all useless operations when we filter on a field - so we know its value in advance.

marcenacp:
That way, we can:

- download parquet 1
- yield examples parquet 1
- download parquet 2
- etc.

Instead of:

- dowloading all parquets
- yielding all examples from all parquets

I also added a non-hermetic test which should timeout if there's a regression on this feature.

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/datasets_test.py
+++ b/python/mlcroissant/mlcroissant/_src/datasets_test.py
@@ -102,22 +102,27 @@ def load_records_and_test_equality(
         f" {record_set_name} --num_records {num_records} --debug --update_output"
         f" {filters_command}`"
     )
-    config = _REPOSITORY_FOLDER / "datasets" / version / dataset_name
-    output_file = config.parent / "output" / f"{record_set_name}.jsonl"
-    with output_file.open("rb") as f:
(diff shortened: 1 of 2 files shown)


pierrot0 (approved):

ccl-core (approved):
Thanks!

python/mlcroissant/mlcroissant/_src/datasets_test.py
```diff
@@ -248,6 +253,15 @@ def test_nonhermetic_loading(version, dataset_name, record_set_name, num_records
         ["huggingface-c4/metadata.json", "data", 1, {"data/variant": "en"}],
         ["huggingface-levanti/metadata.json", "levanti_train", 10, None],
         ["huggingface-open-hermes/metadata.json", "default", 3, None],
+        # This dataset will timeout if the following feature is broken: mlcroissant
```
ccl-core, line 256: Shall we add a meaningful error message somewhere?

marcenacp: I would rather keep it simple with a comment - otherwise it means intercepting pytest's timeout or create a timeout, etc

PR #776, opened 2024-11-29
Fix errors with nested subfields

ccl-core:


ccl-core:
Tested with:

```
python load.py --jsonld   /usr/local/google/home/stanzelc/croissant/datasets/1.0/huggingface-lmms-eval-lite/metadata.json --record_set gqa  --debug --num_records 3 
```

ccl-core:
TODO: add tests

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
```diff
@@ -203,7 +204,9 @@ def call(self, df: pd.DataFrame) -> Iterator[dict[str, Any]]:
             for _, row in df.iterrows():
                 yield dict(row)
             return
-        fields = self._fields()
+        fields = self._get_fields(self.node.fields)
+        if not fields:
+            raise ValueError(f"RecordSet {self.node.uuid} has no fields!")
```
marcenacp, line 209: Is this wanted? We probably want to return an empty dict?

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
```diff
@@ -226,9 +229,13 @@ def _get_result(row):
...
-                    value = [
-                        _cast_value(self.node.ctx, v, field.data_type) for v in value
-                    ]
+                    try:
+                        value = [
+                            _cast_value(self.node.ctx, v, field.data_type)
+                            for v in value
+                        ]
+                    except TypeError:
+                        value = value
```
marcenacp, line 238: What do you think of dropping the _cast_value?

marcenacp (commented):
Thanks for the thorough bug analysis!

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
ccl-core: Are you sure we want to do that for all cases?

PR #779, opened 2024-12-03
Clean code by checking attribute.

marcenacp:


python/mlcroissant/mlcroissant/_src/structure_graph/base_node.py
```diff
@@ -205,14 +207,14 @@ def uuid(self) -> str:
             return self.id
 
     @property
-    def parent(self) -> "Node | None":
-        """Direct parent of the node or None if no parent."""
+    def parent(self) -> Node:
+        """Direct parent of the node."""
         if not self.parents:
-            return None
+            raise ValueError(f"node={self} does not have any parent.")
```
ccl-core, line 213: There are no nodes that can have no parents? What about the Metadata node?

ccl-core (approved):

python/mlcroissant/mlcroissant/_src/structure_graph/base_node.py
marcenacp: Done.

PR #787, opened 2024-12-17
Add Python version requirement to project metadata

amercader:
Although the package requires Python>=3.10 it can still be installed in lower versions of Python because the requirement is not listed in the project metadata, so PyPI can't prevent it.

### DIFF
--- a/python/mlcroissant/pyproject.toml
+++ b/python/mlcroissant/pyproject.toml
@@ -15,7 +15,7 @@ authors = [
   { name = "Pierre Ruyssen" },
   { name = "Prabhant Singh" },
 ]
-
+requires-python = ">= 3.10"
 # pip dependencies of the project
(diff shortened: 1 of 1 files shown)


python/mlcroissant/pyproject.toml
```diff
@@ -15,7 +15,7 @@ authors = [
   { name = "Pierre Ruyssen" },
   { name = "Prabhant Singh" },
 ]
-
+requires-python = ">= 3.10"
```
bact, line 18:
```suggestion
requires-python = ">= 3.10, <=3.12"
```

Support in Python 3.13 maybe limited, for the current codebase, due to the use of aifc and audioop which are deprecated and to be removed in 3.13.

See https://peps.python.org/pep-0594/#deprecated-modules

bact (changes requested):

Issue #796, opened 2025-01-24
Croissant definitions for Titanic are broken


ccl-core:
Current croissant definitions for the Titanic dataset are broken for both 1.0 and 0.8

This happens because `contentUrl` for the `passengers.csv` FileObject points to a url which is no longer valid: https://www.openml.org/data/get_csv/16826755/phpMYEkMl

This causes our GH actions to fail, as the dataset is used to test mlcroissant's generation scripts. The croissant definitions should be updated.

joaquinvanschoren:
Hi, the url was temporarily unavailable due to a cyberattack. It should work again. 

Sorry for the inconvenience. Let me know if you encounter any further issues.

ccl-core:
Thank you a lot for the update Joaquin!
Tests are green now :)

NevadaM:
Hi both - the Titanic example dataset is broken in the HF editor (see image) and I'm curious if it's the same issue? I haven't had the time to reproduce the error in a local build yet, so can't see logs to confirm. @joaquinvanschoren any recent cyberattacks breaking the url? 

If not, I'll maybe PR the editor files to temporarily remove Titanic (and other broken example datasets) from the editor, then work towards a fix.  

![Image](https://github.com/user-attachments/assets/8f672fc7-6239-4633-8cd7-4abd1d8d8aa6)

joaquinvanschoren:
Hmm, the url seems to work fine, and OpenML is fully operational without any known issues: https://www.openml.org/data/get_csv/16826755/phpMYEkMl

Also, if I go to the Titanic dataset on OpenML, the croissant file seems to download fine.
Did you switch to using a different URL? Or is it another issue?

NevadaM:
Ah thanks for getting back. Ok, that's good to know - I've spent some time looking through the editor code and will probably find what the issue is then. Cheers :)

PR #797, opened 2025-01-24
Remove titanic from ci tests as it is broken

ccl-core:
#796


ccl-core:
Not needed anymore as the links are fixed.

PR #801, opened 2025-01-27
state media type for Croissant (includes profile)

### LINKED ISSUE #792: media type / content type / MIME type for Croissant (JSON-LD "profile" parameter)
## tl;dr

Can the Croissant [spec](https://docs.mlcommons.org/croissant/docs/croissant-spec.html) be updated to formally state that the media type for Croissant is `application/ld+json; profile="http://mlcommons.org/croissant/1.0"` (the value from `"conformsTo": "http://mlcommons.org/croissant/1.0"`)?

## Long version

During the excellent presentation @stain gave this week on [Signposting](https://signposting.org) I learned that JSON-LD allows you use a "profile" parameter to help distinguish between formats that are all "application/ld+json" (such as Croissant, CodeMeta, RO-Crate, etc.).

For example, the RO-Crate docs have a section called [RO-Crate JSON-LD Media type](https://www.researchobject.org/ro-crate/specification/1.1/appendix/jsonld.html#ro-crate-json-ld-media-type) that shows the following example (note the use of "profile"):

`HEAD http://example.com/ro-123/ro-crate-metadata.json HTTP/1.1`
`HTTP/1.1 200 OK`
`Content-Type: application/ld+json; profile="http://www.w3.org/ns/json-ld#flattened http://www.w3.org/ns/json-ld#compacted https://w3id.org/ro/crate"`

During the call I asked what I should use for "profile" for Croissant. @benjelloun indicated I should use the value from `"conformsTo": "http://mlcommons.org/croissant/1.0"` like this:

`Content-Type: application/ld+json; profile="http://mlcommons.org/croissant/1.0"`

This makes complete sense and I'll try it but it would be nice to see this mentioned specifically in the spec. I'm happy to create a pull request if you like!

For more on the JSON-LD "profile" parameter, here's a screenshot of some docs from https://www.w3.org/TR/json-ld/#application-ld-json

Screenshot 2025-01-10 at 3 47 13 PM

pdurbin:
- closes #792

### DIFF
--- a/docs/croissant-spec-draft.md
+++ b/docs/croissant-spec-draft.md
@@ -236,6 +236,12 @@ Croissant datasets must declare that they conform to this specification by inclu
 
 Note that while the Croissant specification is versioned, the Croissant namespace above is not, so the constructs within the Croissant vocabulary will keep stable URIs even when the specification version changes.
 
+The media type (content type or MIME type) for Croissant includes a JSON-LD [profile](https://www.w3.org/TR/json-ld/#application-ld-json) to distinguish it from other JSON-LD documents:
+
+```
+application/ld+json; profile="http://mlcommons.org/croissant/1.0"
+```
+
 ### ID and Reference Mechanism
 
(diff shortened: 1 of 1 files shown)


pierrot0:
I am ok with this, adding Omar to have double check.

benjelloun (approved):

PR #807, opened 2025-02-12
Include isArray and arrayShape

ccl-core:
Addresses: #649

Update: the final implementation uses:

```
{
          "@type": "cr:Field",
          "@id": "recordset/fieldname",
          "isArray": true,
          "arrayShape": "1,2,1"
       ...
}
``` 

======

Note that this is a bit different than the original proposal in #649 :

```
{
          "@type": "cr:Field",
          "@id": "recordset/fieldname",
          "isArray": true,
          "arrayShape": [
            1,
            2,
            1
          ]
       ...
}
```

This is a bit difficult for us to process using mlcroissant... When we create the [graph from the jsonld](https://github.com/mlcommons/croissant/blob/d98ba0e31c7d8152c46e5a6d81dc4a017d8b1741/python/mlcroissant/mlcroissant/_src/core/json_ld.py#L210) data, we would loose the order information; moreover, as we create a triple for each element in the `arrayShape`, and triples need to be unique, we would loose dimensions which have the same size.

As an alternative I propose to use a `@list`, so we keep the order info more nicely and we don't loose data:

```
{
          "@type": "cr:Field",
          "@id": "recordset/fieldname",
          "isArray": true,
          "arrayShape": {
            "@list": [
              1,
              2,
              1
            ]
          },
       ...
}
```

WDYT?

In this PR:
* Introduces `isArray` and `arrayShape`
* Includes one example dataset (datasets/1.0/huggingface-baratilab-flow3d/metadata.json) which is a simplified version of https://huggingface.co/datasets/baratilab/Flow3D-V0-Ti-6Al-4V_fast_test (less fields, only one config) which includes the new fields.
* Updates all contexts

Still TODO: Remove `repeated` from all datasets and use `isArray` and `arrayShape` instead.


datasets/1.0/huggingface-squad/metadata.json
```diff
@@ -106,6 +113,11 @@
           "name": "record_set_plain_text/split",
           "description": "Split to which the example belongs to.",
           "dataType": "sc:Text",
+          "references": {
```
marcenacp, line 116: nit: `references` coming after `source` was more readable.

python/mlcroissant/mlcroissant/_src/core/constants.py
```diff
@@ -27,6 +27,7 @@ def ML_COMMONS(ctx) -> rdflib.Namespace:
         return ML_COMMONS_V_1_0
 
 
+ML_COMMONS_ARRAY_SHAPE = lambda ctx: ML_COMMONS(ctx).arrayShape
```
marcenacp, line 30: Should we write a test that the verifier fails if `isArray` is specified with no `arrayShape`? Or is this legit in the specs?

If there's an implicit behaviour that arrayShape defaults to (-1,), we could probably document this

benjelloun:
As we discussed offline, arrayShape can be represented more compactly as a string that contains space separated dimensions, e.g., "1 2 -1 3", where -1 corresponds to a dimension where the length is not specified. Even though this approach is encoding structured information in a string, it's unambiguous and easy to parse, so I think it's acceptable. We already do accept such input in other places, e.g., for bounding boxes.

python/mlcroissant/mlcroissant/_src/core/constants.py
marcenacp: After the discussion with Omar:

- `isArray` without `arrayShape` is valid, and in that case `arrayShape` defaults to (-1,). The documentation should specify this behaviour.
- `arrayShape` without `isArray` should make the verifier fail.

ccl-core: Godo point, done. I'll update the documentation in a follow-up PR, for now I just updated the docstring.

datasets/1.0/huggingface-squad/metadata.json
ccl-core: Yeah I agree... I don't know why it got changed, maybe it started to order those alphabetically? I could change it back manually, shall I do this?

datasets/1.0/huggingface-mnist/metadata.json
```diff
@@ -2,6 +2,7 @@
   "@context": {
     "@language": "en",
     "@vocab": "https://schema.org/",
+    "arrayShape": "cr:arrayShape",
```
marcenacp, line 5: Should all these modifications appear in a new `datasets/1.1` folder?

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/field.py
```diff
@@ -40,6 +40,20 @@ class Field(Node):
 
     JSONLD_TYPE = constants.ML_COMMONS_FIELD_TYPE
 
+    array_shape: list[int] | None = mlc_dataclasses.jsonld_field(
```
marcenacp, line 43: nit: I would have `array_shape` mirror exactly `cr:arrayShape` (i.e., `str | None`). Then, I would have `array_shape_list` be a method of `Field` that has the logic.

```python
def array_shape_list(self) -> Sequence[int]:
  if self.is_array and not self.array_shape is None:  # example: this is not trivial logic
    return (-1,)
  # etc
```

datasets/1.0/huggingface-mnist/metadata.json
marcenacp: (Could be done in another PR)

ccl-core: As discussed offline, added a 1.1 folder with the new dataset. The 1.0 datasets are not cmodified. Will update the old datasets and put the new versions in 1.1 once repeated is deprecated.

ccl-core:
@marcenacp WDYT if I add a "IS_LATEST_VERSION" attr to /usr/local/google/home/stanzelc/croissant/python/mlcroissant/mlcroissant/_src/core/context.py ? Then we could add the new elements to the context in rdf.py, include 1.1 to the versions we use for testing in /usr/local/google/home/stanzelc/croissant/python/mlcroissant/mlcroissant/_src/tests/versions.py

datasets/1.0/huggingface-rag-dataset/metadata.json
```diff
@@ -40,9 +42,43 @@
     "separator": "cr:separator",
     "source": "cr:source",
     "subField": "cr:subField",
-    "transform": "cr:transform"
+    "transform": "cr:transform",
+    "arrayShape": "cr:arrayShape",
+    "isArray": "cr:isArray",
```
marcenacp, line 47: For readability, can we follow the alphabetical order in all `@context`?

(Can be done in another PR)

datasets/1.0/huggingface-squad/metadata.json
marcenacp: An array of strings used to control the order. We shouldn't do it manually but programmatically. But this is not prio at all - so we can probably let it like this if we don't have time.

marcenacp: Resolved.

python/mlcroissant/mlcroissant/_src/core/rdf.py
```diff
@@ -25,6 +25,9 @@ def make_context(ctx=None, **kwargs):
     context = {
         "@language": "en",
         "@vocab": "https://schema.org/",
+        "arrayShape": (
+            "cr:arrayShape" if ctx is not None and ctx.is_latest_version() else None
```
marcenacp, line 29: The condition should be `ctx.version >= 1.1`? Otherwise, when increasing the version, we'll lose cr:arrayShape?

Same below.

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/field.py
```diff
@@ -130,6 +153,11 @@ def __post_init__(self):
         self.assert_has_mandatory_properties(uuid_field)
         self.source.check_source(self.add_error)
         self._standardize_data_types()
+        if self.array_shape and not self.is_array:
+            self.add_error(
+                f"Field {self.uuid} defines `array_shape`, but `is_array` is False."
+            )
+        self.array_shape_tuple
```
marcenacp, line 160: Remove?

marcenacp: ah you probably use this to check that the string is correct (ints separated by `,`)!

marcenacp: Note that this will fail - and not dispatch an error as the other checks.

datasets/1.0/huggingface-rag-dataset/metadata.json
ccl-core: Done.

ccl-core: Done.

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/field.py
ccl-core: Added explicit error and added one more e2e test.

python/mlcroissant/mlcroissant/_src/core/rdf.py
ccl-core: Addressed in https://github.com/mlcommons/croissant/pull/807/commits/369c07f5cd31fd2b64198b7fed4cf81502b007a0

PR #814, opened 2025-02-24
Fix dependencies in the editor and check that dependencies are correctly referenced by Hugging Face.

marcenacp:


editor/deploy_to_hf.sh
```diff
@@ -2,6 +2,12 @@ HF_REPO=/tmp/hf-croissant
 echo "Deleting $HF_REPO..."
 rm -rf ${HF_REPO}
 git clone git@hf.co:spaces/marcenacp/croissant-editor ${HF_REPO}
+streamlit_version_in_hf=$(grep -oP 'sdk_version: \K[0-9]+\.[0-9]+\.[0-9]+' ${HF_REPO}/README.md)
+streamlit_version_in_gh=$(grep -oP 'streamlit\=\=\K[0-9]+\.[0-9]+\.[0-9]+' requirements.txt)
```
NevadaM, line 6: so this means that it is mandatory to have "streamlit==x.x.x" in requirements.txt - we can't have just "streamlit" or "streamlit>=x.x.x"? makes sense for now, I think will be a longevity problem but we'll have to sort it out when we overhaul editor anyway.

editor/requirements.txt
```diff
@@ -6,6 +6,6 @@ pytest
 python-magic
 rdflib
 requests
-streamlit
+streamlit==1.39.0
```
NevadaM, line 9: info - current stable release of streamlit is 1.42.2 , but streamlit-nested-layout is only verified tested on 1.39.0

NevadaM (approved):
thank you Pierre! will need to point out these changes to anyone working on the editor in the future.

editor/requirements.txt
marcenacp: https://github.com/mlcommons/croissant/pull/815

PR #819, opened 2025-02-27
[editor] Added support for displaying and editing `RecordSet` data types and `Field` equivalentProperty

david4096:
Currently those features are not observable or editable in the editor, but they are important to give clearer meaning to the records and fields.

- For now the RecordSet dataTypes are shown in 1 text box with each data type separated by a comma. It could be changed to a button and multiple fields, but that would be a lot more code without necessarily be more readable, let us know how do you feel. Ideally there should be some kind of autocomplete combobox that takes multiples values, with a predefined list of types to select from, and still the possibility to add any types the user wants. But we could not find such component for now, so we went for the simpler solution.
- The `equivalentProperty`  has been added right after the Field dataType in the table on the left side in the Record Sets tab, as well as in the field details on the right side. We can remove it from the left side table if you feel it is too much and should only be shown in the details

Doing so made us realize some confusing things with the editor example, e.g. with the titanic dataset the RecordSets `genders` and `embarkation_ports` have the type https://schema.org/Enumeration but the checkbox "The RecordSet is an enumeration" is not checked. Wouldn't it make sense to be automatically checked in this case?

We did this during a short hackathon at the https://www.swat4ls.org/ conference (semantic web for life sciences).

It is a beginning of answer to this issue: https://github.com/mlcommons/croissant/issues/739 raised by @benjelloun
 
A lot of people in our community (semantic web for life science research) are interested by the features requested in this issue

We followed the official specs definition for `RecordSet` and `Field`: https://docs.mlcommons.org/croissant/docs/croissant-spec.html#field 

We tested the new features using the docker deployment locally and they work as expected. RecordSet Data types and Field equivalentProperty are properly added to the exported JSON-LD file

It was done with @vemonet 

Also note that in the code for `Field` you use camelcase `equivalentProperty` instead of the usual snake case that is used elsewhere, might want to fix this for overall consistency https://github.com/mlcommons/croissant/blob/main/python/mlcroissant/mlcroissant/_src/structure_graph/nodes/field.py#L71

### DIFF
--- a/editor/views/record_sets.py
+++ b/editor/views/record_sets.py
@@ -193,6 +193,8 @@ def _handle_fields_change(record_set_key: int, record_set: RecordSet):
                 field.description = new_value
             elif new_field == FieldDataFrame.DATA_TYPE:
                 field.data_types = [str_to_mlc_data_type(new_value)]
+            elif new_field == FieldDataFrame.EQUIVALENT_PROPERTY:
+                field.equivalentProperty = new_value
     for added_row in result["added_rows"]:
         data_type = str_to_mlc_data_type(added_row.get(FieldDataFrame.DATA_TYPE))
         field = Field(
@@ -200,6 +202,7 @@ def _handle_fields_change(record_set_key: int, record_set: RecordSet):
             name=added_row.get(FieldDataFrame.NAME),
             description=added_row.get(FieldDataFrame.DESCRIPTION),
             data_types=[data_type],
+            equivalentProperty=added_row.get(FieldDataFrame.EQUIVALENT_PROPERTY),
             source=mlc.Source(),
             references=mlc.Source(),
         )
@@ -217,6 +220,7 @@ class FieldDataFrame:
     NAME = "Field name"
     DESCRIPTION = "Field description"
     DATA_TYPE = "Data type"
+    EQUIVALENT_PROPERTY = "Property IRI"
     SOURCE_UID = "Source"
     SOURCE_EXTRACT = "Source extract"
     SOURCE_TRANSFORM = "Source transform"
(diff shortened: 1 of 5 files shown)


david4096:
Here's a quick demo of how tagging can be done using the editor based off this PR

https://github.com/user-attachments/assets/8653e72b-0d91-4d9b-97f8-c2346779b9f1

benjelloun:
This is very cool! I agree that the UI is not ideal, but this is definitely a good start.

Nit: In your demo, you use the same type at the RecordSet level and at the field level. Those would generally not be the same thing. Also, your demo shows a column called "property IRI". Should that be equivalentProperty?

Nit: RecordSet-level data types are an advanced feature of Croissant, so they definitely deserve a tooltip.

david4096:
Thanks @benjelloun ! Looking forward to getting this merged so we can talk about ways to improve adoption of using more careful annotation. For the bio related use cases we identified this as a crucial feature.

> This is very cool! I agree that the UI is not ideal, but this is definitely a good start.
> 
> Nit: In your demo, you use the same type at the RecordSet level and at the field level. Those would generally not be the same thing. Also, your demo shows a column called "property IRI". Should that be equivalentProperty?
> 

For the first one, yes, in practice the URLS would be different, I just meant to demonstrate we can now annotate with URLs at both levels. Sorry for any confusion.

For the second, changed the label to Equivalent property as opposed to property IRI.

> Nit: RecordSet-level data types are an advanced feature of Croissant, so they definitely deserve a tooltip.

Hope I'm interpreting this one correctly but added a little clarification that it expects a URL. Commented inline.

editor/views/record_sets.py
```diff
@@ -273,6 +277,19 @@ def _render_left_panel():
                 on_change=handle_record_set_change,
                 args=(RecordSetEvent.DESCRIPTION, record_set, key),
             )
+            key = f"{prefix}-datatypes"
+            st.text_input(
+                "Data types",
+                placeholder="Provide comma-separated data types for the RecordSet.",
+                help=(
+                    "Records in this set are instances of the corresponding data types"
```
david4096, line 285: @benjelloun Please let me know if something else should go in this tooltip.

NevadaM:
cc @marcenacp for HF space update when this is merged.

I'm going to take a look at the CI Editor test that's failing here - I don't think it's a problem with your additions @david4096

david4096:
I fixed the usage of single quotes. I think the remaining style problems are line length in parts that aren't part of this PR.

benjelloun (approved):

marcenacp:
@david4096 Can you please make the tests green before merging? (Tests are green on main)

You can use [`make`](https://github.com/mlcommons/croissant/blob/main/editor/Makefile) to format your code (or directly VS Code). Thanks!

david4096:
@marcenacp Fixed the formatting issue, current failure appears to be unrelated to this PR from what I can tell?

PR #820, opened 2025-02-28
Include more value types to describe field's precision more accurately

ccl-core:
Include more precise value types.

Also included one example datasets, https://huggingface.co/datasets/pollen-robotics/apple_storage, to showcase the differences between 1.0 (same Croissant as in: https://huggingface.co/api/datasets/pollen-robotics/apple_storage/croissant) and the proposed modifications that would go into 1.1 (see the relative output folders). The fields' precision was taken from: https://datasets-server.huggingface.co/info?dataset=pollen-robotics/apple_storage


marcenacp (approved):
Great, thanks!

PR #829, opened 2025-03-17
Fix bug in repeated fields.

ccl-core:
Fixed a bug with repeated fields.
Also added an example dataset with multiple transforms for the same field.


python/mlcroissant/mlcroissant/_src/structure_graph/nodes/field.py
```diff
@@ -233,3 +233,12 @@ def parent(self) -> Node:
         if parent:
             return parent
         raise ValueError(f"field={self} does not have any parent.")
+
+    @property
+    def is_subfield(self) -> bool:
+        """Whether the field is a subfield of another field."""
+        parent = super().parent
```
marcenacp, line 240: Use `self.parent`?

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/field.py
```diff
@@ -233,3 +233,12 @@ def parent(self) -> Node:
         if parent:
             return parent
         raise ValueError(f"field={self} does not have any parent.")
+
+    @property
+    def is_subfield(self) -> bool:
```
marcenacp, line 238: You don't seem to use this method. Is this still needed?

datasets/1.1/huggingface-wildchat/metadata.json
```diff
@@ -0,0 +1,1538 @@
+{
```
marcenacp, line 1: Those tests are likely not run ([source](https://github.com/mlcommons/croissant/blob/f031d966a44b7f9a7037070f8a5a673dee25cc17/python/mlcroissant/mlcroissant/_src/tests/versions.py#L11)). Can you add `1.1` to the list of tested versions?

marcenacp (commented):
Thanks!

datasets/1.1/huggingface-wildchat/metadata.json
ccl-core: We can't yet, because we do not have 1.1-compliant datasets for all the current datasets we have -- this means that mlcroissant/_src/datasets_test.py::test_static_analysis , mlcroissant/_src/datasets_test.py::test_hermetic_loading , mlcroissant/_src/datasets_test.py::test_raises_when_the_record_set_does_not_exist , ... etc would fail.

We are including 1.1 datasets, however, in mlcroissant/mlcroissant/_src/core/json_ld_test.py

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/field.py
ccl-core: Removed for simplicity

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/field.py
ccl-core: Removed the method as it is unused

datasets/1.1/huggingface-wildchat/metadata.json
marcenacp: Can you add a dedicated test for 1.1? Similar to [this test for 1.0](https://github.com/mlcommons/croissant/blob/f031d966a44b7f9a7037070f8a5a673dee25cc17/python/mlcroissant/mlcroissant/_src/datasets_test.py#L220-L234).

marcenacp (approved):
Thanks! Added 1 more comment, otherwise LGTM.

PR #835, opened 2025-03-27
Add a EncodingFormat for .arff files

ccl-core:


python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
```diff
@@ -8,6 +8,7 @@
 
 from etils import epath
 import pandas as pd
+from scipy.io import arff
```
marcenacp, line 11: Can you lazy load this package?

python/mlcroissant/pyproject.toml
```diff
@@ -20,6 +20,7 @@ authors = [
 # Installed locally with `pip install -e .`
 dependencies = [
   "absl-py",
+  "scipy",
```
marcenacp, line 23: Alphabetical order.

python/mlcroissant/mlcroissant/_src/core/constants.py
```diff
@@ -212,6 +212,7 @@ class EncodingFormat:
     We inherit the wrong naming `encodingFormat` from https://schema.org/encodingFormat.
     """
 
+    ARFF = "application/x-arff"
```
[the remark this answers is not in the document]
ccl-core: Yeah, let me add the notes from the offline conversation.

Note that I made up the new Mime type for it, as I couldn't find it in https://www.iana.org/assignments/media-types/media-types.xhtml

At the end of the day, arff 's correct mime type is indeed `text/plain`.

Alternatively, I guess we could use text/plain and check for the file's extension somewhere in the `_read_file_content` function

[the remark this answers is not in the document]
joaquinvanschoren: `text/plain` should indeed be correct.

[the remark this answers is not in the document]
marcenacp: Options:

- So far we've used the field [sc:encodingFormat](https://schema.org/encodingFormat) to store the MIME type. The description of the field suggests that we could use it to describe the encoding format. Quote:
  ```
  Unregistered or niche encoding and file formats can be indicated instead via the most 
  appropriate URL, e.g. defining Web page or a Wikipedia/Wikidata entry. 
  Supersedes fileFormat.
  ```
  So we could have `"sc:encodingFormat": "https://ml.cms.waikato.ac.nz/weka/arff.html"`
- If we treat [sc:encodingFormat](https://schema.org/encodingFormat) as a MIME type, we have to keep `"sc:encodingFormat": "text/plain"`. But now we need the information that the file is ARFF. We could add [sc:fileFormat](https://schema.org/fileFormat): `"sc:fileFormat": "arff"`.

I prefer the first option, because it corresponds to the description given by schema.org.

Ping @joaquinvanschoren @benjelloun @ccl-core

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
```diff
@@ -89,6 +96,13 @@ def _read_file_content(self, encoding_format: str, file: Path) -> pd.DataFrame:
             download_git_lfs_file(file)
         reading_method = _reading_method(self.node, self.fields)
 
+        if encoding_format == EncodingFormat.ARFF:
+            if scipy is None:
```
marcenacp, line 100: You shouldn't need this as this is already checked for when lazily loading ([source](https://github.com/mlcommons/croissant/blob/5a2e5201ae23d83c992948e7a960dad95f5ea7f3/python/mlcroissant/mlcroissant/_src/core/optional.py#L24-L28)).

python/mlcroissant/mlcroissant/_src/core/constants.py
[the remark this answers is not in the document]
benjelloun: I also like the first option better. Since schema.org is pretty lax about the cardinality of properties, about specifying both as `"sc:encodingFormat": ["text/plain", "https://ml.cms.waikato.ac.nz/weka/arff.html"]` ?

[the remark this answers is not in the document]
marcenacp: Good idea. We'd also need to change [the cardinality](https://github.com/mlcommons/croissant/blob/5a2e5201ae23d83c992948e7a960dad95f5ea7f3/docs/croissant-spec.md?plain=1#L622) from ONE to MANY.

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
ccl-core: I need if for pytype :S I'm doing the same as https://github.com/mlcommons/croissant/blame/5a2e5201ae23d83c992948e7a960dad95f5ea7f3/python/mlcroissant/mlcroissant/_src/torch/torch_adapter/dataloader.py#L72

python/mlcroissant/mlcroissant/_src/core/constants.py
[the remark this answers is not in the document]
ccl-core: Thanks for the comments! I started to do some of the modifications needed. PTAL whenever you have time :)

joaquinvanschoren (approved):
Looks good!

joaquinvanschoren:
Just submitted the review. One sideline comment: from experience, [liac-arff](https://github.com/renatopp/liac-arff) is more robust than scipy's implementation, but I understand that you prefer scipy. liac-arff hasn't been updated in 4 years so maybe not a dependency you want to have. OpenML is switching over from arff to parquet anyway.

python/mlcroissant/mlcroissant/_src/core/constants.py
```diff
@@ -209,9 +209,14 @@ def ML_COMMONS(ctx) -> rdflib.Namespace:
 class EncodingFormat:
     """Supported MIME Types in Croissant.
 
-    We inherit the wrong naming `encodingFormat` from https://schema.org/encodingFormat.
+    Unregistered or niche encoding and file formats can be indicated instead via the most
+    appropriate URL, e.g. defining Web page or a Wikipedia/Wikidata entry.
+    Supersedes fileFormat.
+
+        We inherit the wrong naming `encodingFormat` from https://schema.org/encodingFormat.
```
marcenacp, line 216: Wrong indentation?

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/file_object.py
```diff
@@ -53,9 +53,14 @@ class FileObject(Node):
         input_types=[SDO.Text],
         url=SDO.description,
     )
-    encoding_format: str | None = mlc_dataclasses.jsonld_field(
+    encoding_format: list[str] | None = mlc_dataclasses.jsonld_field(
```
marcenacp, line 56: It would become `encoding_formats` (plural)

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
```diff
@@ -82,13 +89,22 @@ class Read(Operation):
...
             download_git_lfs_file(file)
         reading_method = _reading_method(self.node, self.fields)
+        if EncodingFormat.ARFF in encoding_format:
+            if scipy is None:
+                raise NotImplementedError(INSTALL_MESSAGE)
+
+            data = scipy.io.arff.loadarff(filepath)
+            return pd.DataFrame(data[0])
 
+        encoding_format = encoding_format[0]
```
marcenacp, line 107: You may miss encoding formats if you only take the first? I would rather propose:

```
for encoding_format in encoding_formats:
  ...
    return ...
  ...
```

marcenacp (approved):
Thanks! Just added a few comments.

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
```diff
@@ -82,13 +89,22 @@ class Read(Operation):
...
         filepath = file.filepath
         if is_git_lfs_file(filepath):
             download_git_lfs_file(file)
         reading_method = _reading_method(self.node, self.fields)
+        if EncodingFormat.ARFF in encoding_format:
+            if scipy is None:
+                raise NotImplementedError(INSTALL_MESSAGE)
+
+            data = scipy.io.arff.loadarff(filepath)
+            return pd.DataFrame(data[0])
```
marcenacp, line 105: Maybe raise an explicit error pointing the user to scipy.io.arff if `data` doesn't have the expected shape (i..e, one-element array)

python/mlcroissant/mlcroissant/_src/structure_graph/nodes/file_object.py
ccl-core: Yeah good point!

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
ccl-core: SG, done!

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
ccl-core: SG, done.

PR #849, opened 2025-04-15
Preserve native sample rate, don't auto-resample to 22050

tombagby:
The default in librosa is sr=22050, which forces resampling, it has to be set explicitly to None to preserve the original sample rate. I doubt this is intentional?

If resampling is desired, the field description should support an explicit target sample rate to enable.

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
+++ b/python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
@@ -96,7 +96,7 @@ def _cast_value(ctx: Context, value: Any, data_type: type | term.URIRef | None):
         else:
             raise ValueError(f"Type {type(value)} is not accepted for an image.")
     elif data_type == DataType.AUDIO_OBJECT:
-        output = deps.librosa.load(io.BytesIO(value))
+        output = deps.librosa.load(io.BytesIO(value), sr=None)
         return output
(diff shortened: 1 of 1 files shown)


benjelloun (approved):
Thanks for the fix!

tombagby:
recheck

ccl-core:
Hi @tombagby , thank you for your contribution!

I think you'd need to ru:

```
mlcroissant load --jsonld ../../datasets/0.8/audio_test/metadata.json --record_set records --num_records 2 --debug --update_output
```

```
mlcroissant load --jsonld ../../datasets/1.0/audio_test/metadata.json --record_set records --num_records 2 --debug --update_output
```

to update the test outputs, which should hopefully fix the failing tests.

bact:
@ccl-core in case you have a maintainer right to re-run the test, you can try that as well. I think it should work since we already have #863 merged into the main branch (which is the branch that this PR trying to merge to).

Click on "..." -> "View details"
Screenshot 2568-05-07 at 15 07 34

Click on "Re-run this job"
Screenshot 2568-05-07 at 15 08 40

ccl-core:
Uhm, I thought that the change in this PR would also affect the output records? In any case, I rerun the CI tests and there still are errors: https://github.com/mlcommons/croissant/actions/runs/14850204797/job/41806335710?pr=849

bact:
@ccl-core You're right. It could work with other PRs but not this one.

PR #852, opened 2025-04-16
Fix ExpectedType of containedIn for FileObject in croissant-spec.md

frogcat:
In croissant-spec.md, the ExpectedType of `containedIn` is `Text` for `FileObject` and `Reference` for `FileSet`.
https://github.com/mlcommons/croissant/blob/main/docs/croissant-spec.md?plain=1#L648-L653
https://github.com/mlcommons/croissant/blob/main/docs/croissant-spec.md?plain=1#L711-L716

I believe it would be better to set the ExpectedType to `Reference`.

### DIFF
--- a/docs/croissant-spec.md
+++ b/docs/croissant-spec.md
@@ -647,7 +647,7 @@ In addition, `FileObject` defines the following property:
   </thead>
   <tr>
     <td>containedIn</td>
-    <td><a href="http://schema.org/Text">Text</a></td>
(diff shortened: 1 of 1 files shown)


benjelloun (commented):
Thanks for the fix!

frogcat:
`mlcommons:main` was merged into `frogcat:patch-2` to pass check, see #863 .

PR #855, opened 2025-04-21
Remove extra symbols in RAI Spec

bact:
- Insert a space before "This document describes the responsible AI ..."
- Remove unnecessary `\_ \_`
- Insert a hyphen between "socio demographic"
- Replace "For eg." with "For example"
  - "e.g." is "for example", the current text reads "for for example"
- Remove unnecessary `\`


ccl-core (approved):
Thanks!

PR #848, opened 2025-04-14
Bring croissant-rdf into Croissant repo

david4096:
croissant-rdf is a tool for working with Croissant JSON-LD using RDF. It currently supports converting data from the major dataset providers. It's useful for getting started with creating Knowledge Graphs of datasets and is a good place to demonstrate further semantic features.

* Started at DBCLS Biohackathon Japan 2024
  * https://osf.io/preprints/biohackrxiv/msv7x_v1
* Continued at Elixir Biohackathon 2024
  * https://osf.io/4sgdq_v1/
* Lastly at Semantic Web Applications and Tools for Healthcare and Life Sciences (SWAT4HCLS)
  * Preprint in progress

Currently at https://github.com/david4096/croissant-rdf

This PR will attempt to bring it into this repo.

TODOs:

- [ ]  Get CI tests and code quality checks integrated
- [ ] Normalize docs and license as requested by MLCommons
- [ ] Move over issues from origin repository
- [ ] Other things? let me know!

PR #858, opened 2025-04-22
Further Croissant RDF integration

stefanches7:
According to #848 comments there are further points in the croissant-rdf code integration

- [ ] Use croissant module for getting some dependencies (but keep the package separate)
* see the comment below
- [X] Switch to JAX convention for source tree
- [X] Improve notebook documentation (and fix networkx failure)


stefanches7:
About the common dependencies - I am not quite sure how it should work. Should I somehow reference `mlcroissant\pyproject.toml` in the `croissant-rdf\pyproject.toml`? @marcenacp maybe you would have an insight from browsing the previous PR codebase

cc @david4096

marcenacp:
> About the common dependencies - I am not quite sure how it should work

@stefanches7 I meant that croissant-rdf code would be included in mlcroissant's library. But we didn't go that path. Thanks a lot for the follow up!

croissant-rdf/pyproject.toml
```diff
@@ -25,8 +25,7 @@ dependencies = [
     "kaggle >=1.6.17",
     "openml >=0.15.1",
     "rich >=13.9.4",
```
marcenacp, line 27: nit: Remove rich in favour of tqdm as pointed in the initial PR?

croissant-rdf/tests/test_generate_rdf.py
```diff
@@ -3,7 +3,7 @@
 import pytest
 from rdflib import Graph
 
-from croissant_rdf.providers import HuggingfaceHarvester
+from croissant_rdf import HuggingfaceHarvester
 
 OUTPUT_FILEPATH = "./tests/test_output.ttl"
```
marcenacp, line 8: nit: We should avoid this pattern as pointed out in the initial review. This raises issues when distributing the tests or running them in parallel. Instead we should create [temporary files](https://docs.python.org/3/library/tempfile.html) within the test.

marcenacp: The same comment would apply to all tests in this PR.

marcenacp (approved):
Thanks!

david4096:
This is really great to see, thanks @stefanches7 !

croissant-rdf/pyproject.toml
david4096: We started with tqdm but moved to rich because it's... richer ;) Prefer we keep it otherwise we're sort of thrashing. https://medium.com/pythoneers/from-tqdm-to-rich-my-quest-for-better-progress-bars-afff39985ffc

stefanches7:
> > About the common dependencies - I am not quite sure how it should work
> 
> @stefanches7 I meant that croissant-rdf code would be included in mlcroissant's library. But we didn't go that path. Thanks a lot for the follow up!

Discussed with @david4096 that dependending on the `mlcroissant` package is probably the way to go, especially since we plan to not only dissect CroissantLD into RDF, but also be able to go back. For that, we will use the `mlcroissant` functionality.

marcenacp:
> dependending on the `mlcroissant` package is probably the way to go

If you ever need to do that, let's just merge both packages. Introducing a dependency would be too hard to maintain - for zero return as I underline in the first PR. For now, we can keep them distinct, ping me if we ever need mlcroissant from croissant-rdf.

Thanks again for the good work! Ready to merge once the temporary file issue is solved.

stefanches7:
> ready to merge once the temporary file issue is solved

Thanks to being on Windows the `NamedTemporaryFile` issue popped up, people report all kind of headaches with writing / accessing: https://github.com/bravoserver/bravo/issues/111#issuecomment-838099 (and just see the list of mentions below this comment!).
I've employed a context approach, hope it works well in the long run.

marcenacp (approved):

stefanches7:
Could it be merged?

stefanches7:
How about merging this @marcenacp @ccl-core or does anything prevent this? Thanks

ccl-core:
It seems like the CI tests are still failing? Is it related to this PR?

ccl-core:
Thanks! I'll have a look at the other PR :)
For this PR, I'm ok with squashing and merge.

stefanches7:
I'm not sure I have the power to merge @ccl-core

PR #864, opened 2025-05-07
Split hermetic and non-hermetic tests.

ccl-core:


python/mlcroissant/mlcroissant/_src/datasets_test.py
```diff
@@ -247,74 +246,6 @@ def test_hermetic_loading_1_1(dataset_name, record_set_name, num_records, filter
     )
 
 
-# Non-hermetic test cases (data from the internet).
-@pytest.mark.nonhermetic
```
marcenacp, line 251: Checking: Is the copy/paste right? It seems you forgot e.g. this decorator

marcenacp: This is needed for the [`Makefile`](https://github.com/mlcommons/croissant/blob/7e5889f396de4bd4243ceda7b428320ee9d4bfb8/python/mlcroissant/Makefile#L20)

ccl-core: Ah, right, they are still needed to run `make pytest-fast`, right?

PR #868, opened 2025-05-08
Preserve np.ndarray in field value casting

tombagby:
The current _cast_value will always cast an np.ndarray to a python list. Instead, when the value is an array and the type to cast to is a numpy dtype, use astype and preserve the array.


python/mlcroissant/mlcroissant/_src/operation_graph/operations/field_test.py
```diff
@@ -51,6 +51,21 @@ def test_cast_value(conforms_to, value, data_type, expected):
...
 
+@parametrize_conforms_to()
+@pytest.mark.parametrize(
+    ["value", "data_type", "expected"],
+    [
+        [np.array([1, 2, 3]), DataType.INTEGER, np.array([1, 2, 3])],
+        [np.array([1, 2, 3]), DataType.FLOAT32, np.array([1.0, 2.0, 3.0])],
+    ],
+)
+def test_cast_value_ndarray():
```
marcenacp, line 62:
```python
def test_cast_value_ndarray(conforms_to, value, data_type, expected):
  ...
```

marcenacp (commented):
Thanks!

LGTM if you fix the red tests.

PR #889, opened 2025-06-11
Updated README citation to Neurips paper

JoanGi:
Since the Neurips paper is a more current extension of Croissant, it should be cited instead of the DEEM workshop paper.


ccl-core (approved):
Thanks!

PR #902, opened 2025-07-04
Include annotations in 1.1 specs

ccl-core:
Add info on annotations to the specs. I just copied over from Omar's notes in https://github.com/mlcommons/croissant/issues/737


benjelloun (approved):
Thank you for adding this to the specification!

PR #908, opened 2025-07-14
Update specs on AudioObjects

ccl-core:


benjelloun (commented):
Per discussion in https://github.com/mlcommons/croissant/issues/654 we may want to use XSD data types instead of defining our own. Does that seem feasible to you?

PR #921, opened 2025-07-29
Update croissant-spec-draft.md for Video fields.

ccl-core:


B13rg:
Does the  `VideoObject` type need to be reference in the Appendix 1 Json-LD context or is it covered by the `"sc": "https://schema.org/",` alias?

Otherwise looks good to me!  A native video datatype is a good addition.

Issue #923, opened 2025-07-30
Make Croissant vocabulary URIs resolve to human-readable HTML pages with machine-readable JSON-LD context/definition


kepae:
The Croissant namespace URIs for each expanded property should resolve to human-readable definitions for each term as well as machine-interpretable content embedded in the HTML. This will allow users to easily lookup definitions for croissant-specific fields, just like schema.org vocabulary.

Currently, JSON-LD expansion of a Croissant file yields the IRIs, but they cannot be dereferenced and lead to 404s.
The namespace domain itself is a valid page that describes the project, but the terms and spec all are defined in another space: https://docs.mlcommons.org/croissant/docs/croissant-spec.html

For example, an expansion might yield:

```
"http://mlcommons.org/croissant/field": [
          {
            "@id": "examples/name",
            "@type": [
              "http://mlcommons.org/croissant/Field"
            ],
           "https://schema.org/description": [
              {
                "@language": "en",
                "@value": "The first column contains the name."
              }
            ],
          }
]
```

http://mlcommons.org/croissant/Field yields a 404, whereas https://schema.org/description yields a human-readable HTML page as well as a machine-readable `<script type="application/ld+json">`.

If one were to try and generate pages to live on this domain, we probably need strictly source-of-truth files in this repository that describe all of the terms. From these, HTML pages and their own JSON-LD context payloads can be generated. Or...? Other approaches?

(Please let me know if I'm missing something about where the spec is located/readable. :-))

---

### Edit: After some more discovery, laying out the state of the world.

* The (normative) Croissant spec is served at:
  https://docs.mlcommons.org/croissant/docs/croissant-spec.html.
  * It is generated from [`/docs/croissant-spec.md`](https://github.com/mlcommons/croissant/blob/main/docs/croissant-spec.md).
  * It does not contain embedded machine-readable data (e.g. JSON-LD in `<head>`) to be used to de-reference vocabulary (i.e. this issue).

* The Croissant namespace URL **with a version** redirects to the Croissant spec:
  https://mlcommons.org/croissant/1.0 ->
  https://docs.mlcommons.org/croissant/docs/croissant-spec.html
* **Without a version**, the namespace URL redirects to the MLCommons project page:
  https://mlcommons.org/croissant/ -> https://mlcommons.org/working-groups/data/croissant/

* Other files under [`/docs`](https://github.com/mlcommons/croissant/tree/main/docs/) are also being served. These include:
[...]

PR #929, opened 2025-08-01
1.1 chore: Verify and fix link urls

B13rg:
#923 

I went through the draft and made sure external links worked.  Everything seems to work, with the final urls being the same most of the time.  
This PR:

* Fix a URL typo
* Adjust book reference to use DOI url

Questions:

* Most places reference schemas with `http://`, but some use `https://`.  How should schema urls be referenced in paragraphs and examples?  The links work either way, when followed the http redirects to https.
* There are some 3rd party datasets that are referenced.  They still exist, but may become dead links in the future.  If possible (license permitting) should they be replicated in the repo or placed somewhere like torrent DHT or ipfs?

### DIFF
--- a/docs/croissant-spec-draft.md
+++ b/docs/croissant-spec-draft.md
@@ -84,7 +84,7 @@ Croissant metadata is encoded in JSON-LD.
 {
   "@context": {
     "@language": "en",
-    "@vocab": "https://schema.org/"
+    "@vocab": "http://schema.org/"
   },
   "@type": "sc:Dataset",
   "name": "simple-pass",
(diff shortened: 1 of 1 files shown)


benjelloun (approved):
Thanks for the clean-up!

> Most places reference schemas with http://, but some use https://. How should schema urls be referenced in paragraphs and examples? The links work either way, when followed the http redirects to https. 

For namespace URLs, we use http:// (that's an issue discussed at length in the schema.org community, and you can probably find an older issue about it.)
For regular links, we should use whatever canonical URL is used for the page (so generally https).

> There are some 3rd party datasets that are referenced. They still exist, but may become dead links in the future. If possible (license permitting) should they be replicated in the repo or placed somewhere like torrent DHT or ipfs?

That's a good question. I think there is value in linking to the source of each dataset instead of replicating it. If they disappear, then we should probably change to other examples.

PR #930, opened 2025-08-04
introduction.ipynb: use idiomatic python

wardi:
Some small changes to the introduction.ipynb to make the python examples more idiomatic:

- use consistent 4-space indents
- remove unnecessary terminate file with newline specifically called out in comment
- use range(12) instead of enumerate, if i > 10: break to produce 12 rows


benjelloun:
> does the bot only check the primary email address on a github account?

This may very well be the case. Can you do the registration process with that email address?

wardi:
recheck

PR #936, opened 2025-08-13
Added support for png, jpg and tiff file formats

zigaklun:
Me and my team would like to use Croissant for Earth Observation datasets. This PR introduces support for PNG and TIFF image formats in croissant. Additionally, it modifies the way JPEG files are processed to better align with the requirements of Earth Observation tasks.

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/core/constants.py
+++ b/python/mlcroissant/mlcroissant/_src/core/constants.py
@@ -242,7 +242,8 @@ class EncodingFormat:
     TSV = "text/tab-separated-values"
     TAR = "application/x-tar"
     ZIP = "application/zip"
-
+    PNG = "image/png"
+    TIF = "image/tiff"
 
 class DataType:
     """Data types supported by Croissant."""
--- a/python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
+++ b/python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
@@ -11,6 +11,7 @@ from etils import epath
 import numpy as np
 import pandas as pd
 
+from PIL import Image
 from mlcroissant._src.core.constants import EncodingFormat
 from mlcroissant._src.core.git import download_git_lfs_file
 from mlcroissant._src.core.git import is_git_lfs_file
@@ -181,7 +182,6 @@ class Read(Operation):
                         })
                 elif (
                     encoding_format == EncodingFormat.MP3
-                    or encoding_format == EncodingFormat.JPG
                 ):
                     sampling_rate = _get_sampling_rate(self.node, self.fields)
                     if sampling_rate:
@@ -190,6 +190,11 @@ class Read(Operation):
                         out = deps.librosa.load(file)
                     return pd.DataFrame({
                         FileProperty.content: [out],
+                    })
+                elif encoding_format in {EncodingFormat.TIF, EncodingFormat.JPG, EncodingFormat.PNG}:
+                    img = Image.open(file).convert("RGB")
+                    return pd.DataFrame({
+                        FileProperty.content: [img]
                     })
             raise ValueError(
                 f"None of the provided encoding formats: {encoding_format} for file"


zigaklun:
OK, I have updated the code to use lazy imports.

python/mlcroissant/mlcroissant/_src/core/constants.py
```diff
@@ -242,6 +242,8 @@ class EncodingFormat:
     TSV = "text/tab-separated-values"
     TAR = "application/x-tar"
     ZIP = "application/zip"
+    PNG = "image/png"
+    TIF = "image/tiff"
```
ccl-core, line 246: Can you please order them alphabetically?

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
```diff
@@ -191,6 +188,27 @@ def _read_file_content(
...
+                    try:
+                        arr = deps.tifffile.imread(file)
+                        arr_min, arr_max = arr.min(), arr.max()
+                        arr_norm = (arr - arr_min) / (arr_max - arr_min)
+                        pil_img = deps.PIL_Image.fromarray(
+                            (arr_norm * 255).astype("uint8")
+                        )
+                    except ModuleNotFoundError:
+                        raise NotImplementedError(
+                            "Pillow or tifffile is not installed and is a dependency"
```
ccl-core, line 209: this error message reads strange, can you fix it? You could also give a more actionable fix, like `Please run `pip install ...`

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
```diff
@@ -191,6 +188,30 @@ def _read_file_content(
...
+                        img = deps.PIL_Image.open(file).convert("RGB")
+                    except ModuleNotFoundError:
+                        raise NotImplementedError(
+                            "Missing dependency to read JPG/PNG files. Pillow is not"
+                            " installed. Please, install `pip install pillow`"
+                        )
+                    return pd.DataFrame({FileProperty.content: [img]})
+                elif encoding_format == EncodingFormat.TIF:
+                    try:
+                        arr = deps.tifffile.imread(file)
```
marcenacp, line 202: Is this code repeated from

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
```diff
@@ -95,7 +95,22 @@ def _cast_value(ctx: Context, value: Any, data_type: type | term.URIRef | None):
         if isinstance(value, deps.PIL_Image.Image):
             return value
         elif isinstance(value, bytes):
-            return deps.PIL_Image.open(io.BytesIO(value))
+            try:
+                return deps.PIL_Image.open(io.BytesIO(value))
+            except Exception as e:
+                try:
+                    arr = deps.tifffile.imread(io.BytesIO(value))
```
ccl-core, line 102: Can you add some comments to explain why normalization before converting to a Pillow image? I might be wrong but it seems to me like it could result in a loss of the original data's precision , I assume maybe the users might not always be ok with it ?

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
marcenacp: BTW, why do you need the code in `field.py`?

python/mlcroissant/mlcroissant/_src/core/constants.py
ccl-core: Why only one F ? Shouldn't it be TIFF?

python/mlcroissant/mlcroissant/_src/operation_graph/operations/read.py
zigaklun: I was testing downloading the dataset with mlcroissant and it gave me errors beceuse all imageObject are directly feed into PIL.Image and geotiff files can't be directly converted.

python/mlcroissant/mlcroissant/_src/operation_graph/operations/field.py
zigaklun: I used it to better visualize the data but your right it might be better to leave it out.

python/mlcroissant/mlcroissant/_src/core/constants.py
```diff
@@ -238,9 +238,11 @@ class EncodingFormat:
     MP3 = "audio/mpeg"
     MP4 = "video/mp4"
     PARQUET = "application/x-parquet"
+    PNG = "image/png"
+    TAR = "application/x-tar"
     TEXT = "text/plain"
+    TIF = "image/tiff"
```
ccl-core, line 244: TIFF? or TIF?

ccl-core:
Hi! I'm looking at the errors in the CI tests:

```
  -     'dateCreated': '2024-03-18 00:00:00',
  -     'datePublished': '2024-03-18 00:00:00',
  +     'dateCreated': datetime.datetime(2024, 3, 18, 0, 0),
  +     'datePublished': datetime.datetime(2024, 3, 18, 0, 0),
```

this seems to point to some bug where the datetime fields are not correctly parsed back into JSON?

For what I can see from a quick search in [mlcroissant](https://github.com/search?q=repo%3Amlcommons%2Fcroissant+datePublished&type=code&p=2) it seems like `dateCreated` and `datePublished` include only the date and not the time, e.g. https://github.com/mlcommons/croissant/blob/afa0e4bec8797b46afb42b9fc6ae122c4be61874/python/mlcroissant/mlcroissant/_src/tests/graphs/1.1/multilingual_fields/metadata.json#L58

However, I feel like we should be able to support both.

ccl-core:
Yeah, it seems like a bug... To unblock you, you could just delete those two lines from the croissant and create a bug for this under `Issues` ? Then the tests would become green and we could merge.

By the way, were you able to successfully load the dataset using the `load.py` script? Could you also commit the `output` folder it creates, just for a couple of examples?

PR #947, opened 2025-09-17
Updated search_datasets tool using gemini search + the default (jetty)

SubhadityaMukherjee:
- Redefined the search_datasts tool while using gemini to also do a "google search" using gemini and not just look at the dataset index when doing a search. 
This can be tested using "eclair-client --tool search-datasets --query "tree cover datasets" --use-gemini" after running eclair-server and gemini-cli (following the current documentation)

- Also updated the required packages so we can use 'uv' as the venv/package manager.


SubhadityaMukherjee:
recheck

SubhadityaMukherjee:
recheck

SubhadityaMukherjee:
@joaquinvanschoren can you check if this works now and if it can be merged? If so, I can proceed with the other parts of the project as we discussed

PR #948, opened 2025-09-19
Fix cyclic dependencies in operation graph for field.references

Neo3333:
More details in https://github.com/mlcommons/croissant/issues/904.


ccl-core:
Hi @Neo3333 , thank you for having a look into this! Could you please take a look at the failing tests?

Issue #876, opened 2025-05-21
Convert all 1.0 example datasets to 1.1


ccl-core:
We have a rich set of datasets conforming to Croissant 1.0
We should convert them to the 1.1 specs, and store them in the 1.1 folder to be used as examples for users.

ccl-core:
Hi @SujataSaurabh , thank you for this message!

Exactly, once we have included and implemented all modifications we want to have in 1.1, we can run the migration script in `python/mlcroissant/mlcroissant/scripts/migrations/migrate.py` to migrate the existing datasets in the 1.0 folder. The process should not be manual, because it would be error prone.
Thank you!

PR #952, opened 2025-10-05
Catch up the v1.1 dataset files to match the latest context in rdf.py

SujataSaurabh:
This change should not update anything else except context, and may reorder a few things. This is a first step towards #876


SujataSaurabh:
Sorry, created with wrong github user, will recreate and submit it again

PR #953, opened 2025-10-05
Catch up the v1.1 dataset files to match the latest context in rdf.py

SujataSaurabh:
This change should not update anything else except context, and may reorder a few things. This is a first step towards #876

### DIFF
--- a/datasets/1.1/huggingface-pollen-robotics-apple-storage/metadata.json
+++ b/datasets/1.1/huggingface-pollen-robotics-apple-storage/metadata.json
@@ -2,22 +2,24 @@
   "@context": {
     "@language": "en",
     "@vocab": "https://schema.org/",
-    "arrayShape": "cr:arrayShape",
     "citeAs": "cr:citeAs",
     "column": "cr:column",
     "conformsTo": "dct:conformsTo",
     "cr": "http://mlcommons.org/croissant/",
+    "rai": "http://mlcommons.org/croissant/RAI/",
     "data": {
       "@id": "cr:data",
       "@type": "@json"
     },
-    "dataBiases": "cr:dataBiases",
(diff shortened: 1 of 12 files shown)


datasets/1.1/huggingface-pollen-robotics-apple-storage/metadata.json
```diff
@@ -112,46 +143,41 @@
...
-              "@id": "default_splits/split_name"
-            }
           }
         },
         {
           "@type": "cr:Field",
           "@id": "default/observation.state",
           "name": "default/observation.state",
           "description": "Column 'observation.state' from the Hugging Face parquet file.",
+          "arrayShape": "19",
```
benjelloun, line 153: Nit: I think the most meaningful order is dataType, isArray, arrayShape.

datasets/1.1/huggingface-recipe_RL_data_roberta-base/metadata.json
```diff
@@ -117,21 +139,16 @@
...
-              "@id": "default_splits/split_name"
-            }
           }
         },
         {
           "@type": "cr:Field",
           "@id": "default/steps",
           "name": "default/steps",
           "description": "Column 'steps' from the Hugging Face parquet file.",
+          "arrayShape": "-1,-1",
```
benjelloun, line 149: ditto

ccl-core:
Hi @SujataSaurabh , thank you for your contribution!
I think it would be nice to do all the changes (context, syntax, ...) in one pass, and to use the migration scripts to do that (so that we keep documentation of everything).

In any case, I think it would make sense to wait to update the 1.1 datasets until we are sure that nothing else is going to change in the 1.1 specs...

PR #954, opened 2025-10-05
Copy huggingface-squad dataset from v1.0 to v1.1 with no changes.

SujataSaurabh:
In the subsequent pull request, I plan to write the migration script, and update the files. That way, we can clearly see the diffs as a result of migration script. This is being done for #876 .


benjelloun (commented):
Would it make sense to make the syntax adjustments for 1.1 in the same pull request?

PR #966, opened 2025-11-05
spec: add multilingual example, context

wardi:
A small example at the end of the Format Example section demonstrating how to represent multilingual values as implemented in https://github.com/mlcommons/croissant/pull/932

Also update the context Appendix to allow parsing name and description fields in this format in JSON-LD.


benjelloun (commented):
Sorry for the slow turn-around!

PR #969, opened 2025-11-24
A few more changes to the 1.1 spec

benjelloun:


docs/croissant-spec-draft.md
```diff
@@ -481,142 +481,6 @@ These [schema.org](http://schema.org) properties are recommended for every Crois
   </tr>
 </table>
 
-#### Using external vocabularies
```
ccl-core: Do we want to remove this?

Issue #920, opened 2025-07-29
Improve support for manifest files (Common crawl example)


benjelloun:
From email thread:

I think we can represent CC data by treating the paths.gz as a manifest that "contains" the fileset. This is described [vaguely](https://docs.mlcommons.org/croissant/docs/croissant-spec.html#fileset) in the 1.0 spec: 

"A FileSet is a set of files located in a container, which can be an archive FileObject or a "manifest" file."

We should improve the description in the 1.1 spec, and give an example

You can extend the description you have created by adding a "containedIn" relationship between the FileObject and the corresponding FileSet:

```
"distribution": [
    {
      "@type": "cr:FileObject",
      "@id": "warc.paths.gz",
       ...
    },
    {
      "@type": "cr:FileSet",
      "@id": "warc_paths",
      "containedIn": "warc.paths.gz",
      ...
    },
    {
      "@type": "cr:FileObject",
      "@id": "wat.paths.gz",
       ...
    },
    {
      "@type": "cr:FileSet",
      "@id": "wat_paths",
      "containedIn": "wat.paths.gz",
      ...
    },
    ...
  ]
```

We are stretching things a bit by having the contained be both an gz archive and a manifest, but that seems reasonable to me. We could add some syntax to make things more explicit if needed.

benjelloun:
I'm actually making FileSet and FileObject both extend from sc:DataDownload, which also fixes another problem: making the "distribution" field valid for schema.org (#725).

In my example above, the type of the `containedIn` property on the `FileSet` is `DataSource`. This is what allows us to provide values for `fileObject` and `transform`. Maybe I should have written it more clearly as:

```
{
  "@type": "cr:FileObject",
  "@id": "manifest.zip",
  "contentUrl": "http://example.com/manifest.zip",
  "encodingFormat": "application/zip"
},
{
  "@type": "cr:FileSet",
  "@id": "my-files",
  "containedIn": {
    "@type": "cr:DataSource",
    "fileObject": { "@id": "manifest.zip" },
    "transform": { "unArchive": true, "readLines": true }
  }
}
```

In general `DataSource` is not at the same level as `FileObject` or `FileSet`. It is used inside the `Field` of a `RecordSet` to point to its data source via the `source` property. We are here extending its use to the `containedIn` property of `FileSet` (and `FileObject` if that makes sense.)

This also allows you to do things like extract a list of files from a column of a CSV file for instance.

Does this make sense to you?

Best,
Omar

[...]

PR #988, opened 2025-12-10
Add 1.1 dataset: commoncrawl CC-MAIN-2025-43

handecelikkanat:
Discussed in [issue](https://github.com/mlcommons/croissant/issues/920#issuecomment-3613172113)

@benjelloun @mqzhou-dev 

- Currently only `metadata.json` is added. I still mark as ready to merge in case this is already usable for you, but please make sure it doesnt break your system if `output` isnt there yet.
- I need sometime to think how to represent our dataset in your `output` format. Seems nontrivial for our data, hopefully other examples will help.
    - Let me know if you `output` urgently (for testing 1.1) as well, rather than only for documentation purposes.
    - Otherwise Ill aim for figuring out by next week, and ask for guidance if I cant figure out by then.
- I left context to use `https://schema.org/`, rather than `http`, so this file can act as a test to support https version also. lmk if you prefer me to change.

### DIFF
--- a/datasets/1.1/commoncrawl-CC-MAIN-2025-43-draft/metadata.json
+++ b/datasets/1.1/commoncrawl-CC-MAIN-2025-43-draft/metadata.json
@@ -1,0 +1,251 @@
+{
+  "@context": {
+    "@language": "en",
+    "@vocab": "https://schema.org/",
+    "sc": "https://schema.org/",
+    "cr": "http://mlcommons.org/croissant/",
+    "rai": "http://mlcommons.org/croissant/RAI/",
+    "dct": "http://purl.org/dc/terms/",
+    "prov": "http://www.w3.org/ns/prov#",
+    "citeAs": "cr:citeAs",
+    "column": "cr:column",
+    "conformsTo": "dct:conformsTo",
+    "data": {
+      "@id": "cr:data",
+      "@type": "@json"
+    },
+    "dataType": {
+      "@id": "cr:dataType",
+      "@type": "@vocab"
+    },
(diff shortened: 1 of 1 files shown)


handecelikkanat:
> @handecelikkanat
> 
> As for the integration test for #920, my plan is to add the test `["commoncrawl-CC-MAIN-2025-43-draft/metadata.json", "default", 1, None],`" for Croissant 1.1 only under `test_nonhermetic_loading_1_1`.
> 
> I noticed that there is no output file for `commoncrawl-CC-MAIN-2025-43-draft`. My understanding is that it is needed for the test to compare against. Could you add one please? Thanks!

@mqzhou-dev , @benjelloun thinks we might not be using an `output` data since we dont have any RecordSets in this fileset. Please see the lengthy discussion on the potential contents of `output`, starting from [here](https://github.com/mlcommons/croissant/issues/920#issuecomment-3636283810).
- Omar's idea was that we will not have an `output` folder at all.
- But he asked the ccl-core team for their opinion.
- Currently I dont have more instructions from ccl-core about this, so I left it unadded - if they instruct me to add some actual data, they also need to detail what to be added - our data is really unconventional so I do not know.
- How about adding an empty output folder in the meanwhile? Would that be enough for test?


> As for the integration test for https://github.com/mlcommons/croissant/issues/920, my plan is to add the test ["commoncrawl-CC-MAIN-2025-43-draft/metadata.json", "default", 1, None]," for Croissant 1.1 only under test_nonhermetic_loading_1_1.

I will leave this to your knowledge, I do not understand the croissant code enough to have an opinion :pray: Thank you!

PR #990, opened 2025-12-19
Add simple stats section to the Croissant 1.1 spec.

benjelloun:


benjelloun:
> @benjelloun To make it clear that both the short and long version of the "dataType" is supported at the Field level, I would add a couple of entries as shown below (min and max).

Added one for max.

> 
> Could you add an inline comment to the JSON (or you need this to be valid JSON?). YAML could be a better way make this human readable.

Should be valid JSON

> 
> You may also add a link and recommend the use of the [DDI Controlled Vocabulary](https://vocabularies.cessda.eu/vocabulary/SummaryStatisticType).
> 

Done, but I used the URL I had before, which seems more authoritative: 
http://rdf-vocabulary.ddialliance.org/cv/SummaryStatisticType/2.1.2/

> ```json
>           "annotation": [
>             {
>               "@id": "person/age/min",
>               "value": 16,
>               "dataType": "ddi-stats:a1d0ec6"
>             },
>             {
>               "@id": "person/age/max",
>               "value": 75,
>               "dataType": "ddi-stats:8321e79"
>             },
>             {
>               "@id": "person/age/mean",
>               "value": 42.51,
>               "dataType": {
>                 "@type": "sc:DefinedTerm",
>                 "termCode": "ArithmeticMean",
>                 "name": "Arithmetic Mean",
>                 "@id": "ddi-stats:7975ed0",
>                 "inDefinedTermSet": "http://rdf-vocabulary.ddialliance.org/cv/SummaryStatisticType/2.1.2/"
>               }
>             }
>            ] 
> ```

PR #1009, opened 2026-02-04
revise licensing section in README.md

luisoala:
- move the apache license bit from readme to bottom of page
- clarify the scope of the two licenses to avoid confusion


luisoala:
cc @benjelloun feel free to rephrase

id just make sure

- licensing is explained in one place
- scope for each is explained to avoid confusion

benjelloun (approved):
Thanks a lot Luis for this update!

PR #1016, opened 2026-03-23
Create tasks sub-folder with .ttl schema.org and SHACL files + basic python validator

leobianco:
This pull request creates the tasks sub-folder to the Croissant repository, providing (1) an extension of schema.org in croissant-tasks.ttl to define metadata for machine learning tasks, (2) SHACL shapes in croissant-tasks-shapes.ttl enforcing the format compliance, (3) A validator.py script + a simple test suite targeting both valid and invalid JSON-LD examples.


leobianco:
Hi,

The workflow is failing at the notebook tests, it seems that it needs to be refactored? Here is an AI-generated explanation + potential fix:

**The Problem**
If you look closely at the logs, the installation step fails silently (or rather, pip prints an error message but the notebook continues executing) before it reaches the cell that imports etils:

1. Permission Denied for APT:

`E: Could not open lock file /var/lib/dpkg/lock-frontend - open (13: Permission denied)`

The notebook is trying to run !apt-get install without sudo, but the GitHub actions runner requires sudo to install system packages.

2. Invalid pip syntax:

```
error: invalid-egg-fragment
× The 'mlcroissant[dev]' egg fragment is invalid
```

Modern versions of pip dropped support for the old #egg=name[extra] syntax when installing from git repositories. Because this failed, the dependencies (including etils) were never actually installed.

3. The Final Crash:

`ModuleNotFoundError: No module named 'etils'
`
Because the pip install in the previous cell failed, Python crashes as soon as it tries to import etils.

**The Solution**
You need to edit the file recipes/bounding-boxes.ipynb (and potentially any other .ipynb notebook files in the recipes/ folder that have this same installation cell).

Open the notebook, find the very first code block that looks like this:

```
# Install mlcroissant from the source
!apt-get install -y python3-dev graphviz libgraphviz-dev pkg-config
!pip install "git+https://github.com/${GITHUB_REPOSITORY:-mlcommons/croissant}.git@${GITHUB_HEAD_REF:-main}#subdirectory=python/mlcroissant&egg=mlcroissant[dev]"
```

And replace the two commands with correctly formatted versions:

```
# Install mlcroissant from the source
!sudo apt-get install -y python3-dev graphviz libgraphviz-dev pkg-config
!pip install "mlcroissant[dev] @ git+https://github.com/${GITHUB_REPOSITORY:-mlcommons/croissant}.git@${GITHUB_HEAD_REF:-main}#subdirectory=python/mlcroissant"
```

What changed:
1. Added sudo in front of apt-get so it has permissions on the test runner.
2. Changed the pip syntax to the modern "Direct URL" syntax (name[extra] @ url) instead of appending &egg=... to the end of the URL.

tasks/testdata/invalid_problem_no_spec.jsonld
benjelloun, line 1: Nit: Maybe rename to invalid_problem_no_spec to explain why this problem is invalid? (It took me a while to understand the reason for it to be invalid.)

tasks/testdata/invalid_solution_no_is_based_on.jsonld
benjelloun, line 1: Nit: Consider renaming to invalid_solution_no_problem to clarify what makes this solution invalid.

tasks/croissant-tasks-shapes.ttl
```diff
@@ -0,0 +1,144 @@
...
+    sh:path croissant:execution ;
+    sh:class croissant:ExecutionInfo ;
+    sh:message "croissant:execution must point to an ExecutionInfo."
+  ] ;
+  sh:property [
+    sh:path croissant:evaluation ;
+    sh:class croissant:EvaluationTask ;
+    sh:message "croissant:evaluation must point to an EvaluationTask."
+  ] ;  
+  sh:property [
```
benjelloun, line 52: You may want to add a rule to check that a Croissant task has either an implementation or one or more subtasks.

tasks/croissant-tasks-shapes.ttl
```diff
@@ -0,0 +1,144 @@
...
+
+# -----------------------------------------------------------------------------
+# TaskProblem Shape
+# -----------------------------------------------------------------------------
+croissant:TaskProblemShape
+  a sh:NodeShape ;
+  sh:targetClass croissant:TaskProblem ;
+  # A TaskProblem must have at least one property that is a spec class
+  sh:property [
+    sh:or (
```
benjelloun, line 66: Why not the execution or evaluation?

tasks/croissant-tasks.ttl
```diff
@@ -0,0 +1,101 @@
...
+croissant:TaskSolution a rdf:Class ;
+  rdfs:label "TaskSolution" ;
+  rdfs:comment "A TaskSolution represents a specific, concrete answer to a TaskProblem. It provides the actual values for the components that were marked as Specs in the problem definition. A TaskSolution inherits all properties from cr:Task to describe the concrete components of the solution." ;
+  rdfs:subClassOf croissant:Task .
+
+# Task properties
+
+croissant:input a rdf:Property ;
+  rdfs:label "input" ;
+  rdfs:comment "The primary data used as input for the task, typically provided as a Croissant dataset. This field can be repeated in case of multiple input datasets." ;
```
benjelloun: Nit: Consider rephrasing the end of the comment, as inputs don't have to be datasets.

tasks/croissant-tasks.ttl
```diff
@@ -0,0 +1,101 @@
...
+  rdfs:label "execution" ;
+  rdfs:comment "Information about the execution of the task." ;
+  schema:domainIncludes croissant:Task ;  
+  schema:rangeIncludes croissant:ExecutionInfo .
+
+croissant:evaluation a rdf:Property ;
+  rdfs:label "evaluation" ;
+  rdfs:comment "The evaluation of the task, represented as another task." ;
+  schema:domainIncludes croissant:Task ;  
+  schema:rangeIncludes croissant:EvaluationTask .
```
benjelloun: I don't see croissant:EvaluationTask defined in this ttl file.

tasks/croissant-tasks.ttl
```diff
@@ -0,0 +1,101 @@
...
+
+croissant:subTask a rdf:Property ;
+  rdfs:label "subTask" ;
+  rdfs:comment "One or more subtasks that are part of the definition of this task." ;
+  schema:domainIncludes croissant:Task ;  
+  schema:rangeIncludes croissant:Task .
+
+# Reusing schema:isBasedOn to link a TaskSolution to the TaskProblem it solves.
+schema:isBasedOn schema:domainIncludes croissant:TaskSolution ;
```
benjelloun, line 97: I have second thoughts about reusing isBasedOn vs defining a specific property in the vocab.

benjelloun (approved):
Very nice! Sorry for the late review.

tasks/croissant-tasks-shapes.ttl
leobianco: I think it makes sense to check this for a TaskSolution, adding it on the next commit.

tasks/croissant-tasks-shapes.ttl
leobianco: Will add the execution spec option to the execution field of the TaskProblem, as it might be filled later for a particular TaskSolution execution.

As for the evaluation, intuitively, at least for benchmarks, evaluation should be concrete on the TaskProblem, otherwise each solution could specify its own evaluation metrics?

tasks/croissant-tasks.ttl
leobianco: This is because it is not defined in the Croissant Tasks doc. I will do this.

tasks/croissant-tasks.ttl
leobianco: Maybe we can leave this as is for the first version and refactor later for a better solution?
Or, let me know your alternative ideas, if they are simple enough I can refactor it on time.

tasks/croissant-tasks-shapes.ttl
benjelloun: Agree for benchmarks, but with a bit of imagination, I could imagine some other types of tasks where the evaluation is not specified by the problem...

PR #1017, opened 2026-04-08
Create MMLU example

leobianco:
Created a folder called `benchmark_examples` and added the MMLU benchmark example there. The goal is to demonstrate the expressivity of Croissant Tasks, that is, it can describe a range of different benchmarks. Also, writing examples helps one to develop the spec, so I have updated the Turtle files accordingly.


luisoala:
@leobianco i put the absence bench with two sample runs here https://github.com/mlcommons/croissant/pull/1025, it is based on your mmlu_example branch, run 1 as of yesterday morning, run 2 as of today

i think we can declutter a lot of the examples once they stabilize but i included most of the traces for benefit of our analysis and debugging

i also decided to separate skill outputs into subdir for readability

i might not be able to join the meeting today but we can coordinate async here or on chat for next steps

peymanvahidi:
@leobianco This PR can be merged (#1026) , but I couldn't change the reviewer, could you check it out.

luisoala:
@leobianco updated results traces on absencebench with sanitization on pii for double blind are in this pr https://github.com/mlcommons/croissant/pull/1029

like i said, i think we should eventually split out the experimental artifacts and traces from the pedagogical outputs (stable problem and solution files) to keep main branch lean and mean

PR #1019, opened 2026-04-16
Fix unit tests

ccl-core:
Starting from pip 23.1 the `#egg=pkg_name` syntax used in our notebooks got deprecated.
This PR updates the syntax and moves the common system dependencies to the github action workflow.


benjelloun (approved):
Thanks for the fix!

PR #1021, opened 2026-04-17
Add Croissant visualizer command and template

benjelloun:
This adds a `visualize` command to the mlcroissant CLI that generates an HTML visualization of a Croissant dataset, including metadata, resources (with previews), and record sets.

This also automatically generates an "index.html" file for each dataset in our repository, and updates it when needed.


ccl-core (approved):

ccl-core:
Thank you so much @benjelloun for the PR! This is great :)

Quick question -- I know we already rely on the visualize.yml workflow and visualize_all.py to catch regressions by running the tool on all datasets, but I was wondering whether it might make sense to add tests for edge cases?

benjelloun:
Per @marcenacp Refactored the implementation to move the rendering to a js library, and factored out the CSS as well. This makes the resulting HTML much more compact.

PR #1031, opened 2026-05-19
Fix custom JSON-LD type-checking compatibility for Python 3.14 upgrade

mqzhou-dev:
### Motivation
This PR prepares Python 3.14 upgrade by moving custom JSON-LD type-checking validation from runtime (inside the jsonld_fields generator) to class definition/import time (inside the @mlc_dataclasses.dataclass decorator).

### Background Context
Python 3.14 introduces deferred/lazy evaluation of type annotations ([PEP 649](https://peps.python.org/pep-0649/)).

In mlcroissant, type-checking validation was previously executed dynamically at runtime when traversing fields. If a downstream user or test suite dynamically mocks a type globally (for instance, patching datetime.datetime to return a custom subclass like MockDatetime during unit tests), then resolving type annotations dynamically at runtime will point to the mocked class MockDatetime, while the static class fields established during module import time still point to the original datetime.datetime. This mismatch would result in a strict type-checking TypeError at runtime. (e.g. TypeError: Field "Metadata.date_created" should have type MockDatetime | None. Got datetime.datetime | None)

Moving this check to class definition time ensures all type annotations are evaluated immediately when the module is imported/loaded, making the library robust against runtime type mocking/patching (e.g., for testing components like croissant_pipelines_test internally) and enforcing a clean, fail-fast architecture for static class definitions.

### DIFF
--- a/python/mlcroissant/mlcroissant/_src/core/dataclasses.py
+++ b/python/mlcroissant/mlcroissant/_src/core/dataclasses.py
@@ -192,7 +192,6 @@ def jsonld_fields(cls_or_instance) -> Iterator[JsonldField]:
                 continue
         metadata = cast(Metadata, field.metadata)
         if metadata:
-            _check_types(cls_or_instance, field, metadata)
             yield JsonldField(
                 default=field.default,
                 default_factory=field.default_factory,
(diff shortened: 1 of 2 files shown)


benjelloun (approved):
Can you try to fix the failing tests? Coding agents generally do well at those if you pass them the error message (or let them run the tests themselves.)

ccl-core (approved):

mqzhou-dev:
> Can you try to fix the failing tests? Coding agents generally do well at those if you pass them the error message (or let them run the tests themselves.)

Yes! I have fixed the broken tests in #1032 and #1034. (The broken tests are unrelated to this change)

PR #1032, opened 2026-05-21
Fix MyPy type annotation errors (15 errors → 0)

mqzhou-dev:
## Motivation

  The **MyPy** CI check is currently failing on open PRs — notably #1031 — with
  15 type errors. These errors are **not introduced by any PR's changes**; they
  are pre-existing errors in `main` surfaced by **dependency drift**.

  This PR fixes those 15 errors so the MyPy job goes green again, unblocking
  #1031 (and any other open PR) from merging.

  ## Why `main` shows a green CI badge but a fresh MyPy run fails
This is probably because that the type-checking toolchain is unpinned so that a fresh CI run therefore installs newer releases whose stricter / changed type signatures flag code that previously type-checked clean.

  So running `make mypy` locally today, or re-running CI on any open PR today,
  installs today's toolchain and fails.
  
 ## Next step 
I will work on the other CI failure due to `flores-200` failure in the subsequent CL before merging #1031


ccl-core (approved):

benjelloun (approved):
Thanks!

Issue #1030, opened 2026-05-14
Joining fields of json files leads to `ValueError`


mgrub:
Dear croissant dev-team,

I am really excited about the croissant data description framework. While getting myself familiar with it, I encountered an issue while trying to "join two json files".

In detail, I have two json-files which contain different properties of the same records. In each json file, a unique id is available to associate the records in both files with eachother. So the natural choice for me was to use a join-operation activated by the "references"-statement in the croissant metadata.

I have set up a "toy example" metadata which is quite close to the thing that I actually want to achieve. The full files are attached to this post.

My data files are:

`fileA.json`

```json
[
    {
        "id": "5dc0",
        "prop_A": "red"
    },
    {
        "id": "d5c0",
        "prop_A": "red"
    },
    {
        "id": "5dc4",
        "prop_A": "green"
    },
    {
        "id": "1de8",
        "prop_A": "green"
    },
    {
        "id": "76c0",
        "prop_A": "green"
    }
]
```

`fileB.json`

```json
[
    {
        "id": "5dc0",
        "prop_B": "purple"
    },
    {
        "id": "d5c0",
        "prop_B": "orange"
    },
    {
        "id": "5dc4",
        "prop_B": "black"
    },
    {
        "id": "1de8",
        "prop_B": "blue"
    },
    {
        "id": "76c0",
        "prop_B": "cyan"
    }
]
```

The relevant sections of the croissant metadata are:

The file objects:

```json
  "distribution": [
    {
      "@type": "cr:FileObject",
      "@id": "file_A",
      "name": "file A",
      "contentUrl": "fileA.json",
      "encodingFormat": "application/json",
      "md5": "af557c56f0c59fe926d4812d9dcaed68"
    },
    {
      "@type": "cr:FileObject",
      "@id": "file_B",
      "name": "file B",
      "contentUrl": "fileB.json",
      "encodingFormat": "application/json",
      "md5": "cdaf498a9fac4bcb2e4de3244bb1dd03"
    }
  ]
```

And the records:

```json
  "recordSet": [
    {
      "@type": "cr:RecordSet",
      "@id": "thing_1",
      "key": {
        "@id": "thing_1/hash"
      },
      "field": [
        {
          "@type": "cr:Field",
          "@id": "thing_1/hash",
          "description": "foo bar",
          "dataType": "sc:Text",
          "source": {
            "fileObject": {
              "@id": "file_A"
            },
            "extract": {
              "jsonPath": "$[*].id"
            }
[...]

PR #1035, opened 2026-05-28
Fix JSON-to-JSON join crash from recursive_successors field leak (#1030)

srchilukoori:
## Problem

Joining two JSON FileObjects via `references` crashes with
`ValueError: All arrays must be of the same length` during the
`Read` operation. Reported in #1030.

## Root cause

`_add_operations_for_file_object` in `operation_graph/graph.py`
collects fields for a `Read` operation using `node.recursive_successors`
without filtering. When a field in RecordSet B sources from a field in
RecordSet A (the join reference), the structure graph creates a
cross-RecordSet edge. `recursive_successors` follows this edge and
collects fields from RecordSet B — fields that belong to a different
FileObject. The `Read` operation then tries to extract their jsonPaths
from the wrong JSON file, producing empty arrays that crash
`pd.DataFrame`.

A secondary issue: after fixing the crash, the `Join` merge produces
all-NaN results because `ReadFields` casts `sc:Text` to `bytes` (per
`EXPECTED_DATA_TYPES`), while raw `Read` output keeps `str`. The
pandas merge finds no type-equal matches.

## Fix

1. **`operation_graph/graph.py`** — Filter `recursive_successors` in
   `_add_operations_for_file_object` to only include fields whose
   `source.file_object` (v1.0) or `source.distribution` (v0.8) matches
   the current FileObject.

2. **`operation_graph/operations/join.py`** — Normalize merge key
   columns (decode `bytes` → `str`) before the pandas merge so both
   sides have a common type.

## Testing

Added `json-join` hermetic test fixture (0.8 + 1.0) with two JSON
FileObjects joined via `references`. The fixture covers:

- 1:1 key match
- Duplicate keys (row multiplication)
- Zero-match key (left join → null)

352/352 hermetic tests pass. No regressions in existing join tests
(`simple-join`, `coco2014-mini`).

## Scope

`_add_operations_for_git` and `_add_operations_for_local_file_sets`
have the same unfiltered `recursive_successors` pattern but are not
fixed here — no hermetic test fixtures exist for those paths with
joins. Will file a follow-up issue.


srchilukoori:
@mlcommons/wg-croissant need your review, please.

datasets/0.8/json-join/data/items.json
benjelloun, line 1: I don't think we need a 0.8 version of this dataset.

benjelloun (approved):
Thanks for the fix!

datasets/0.8/json-join/data/items.json
srchilukoori: I agree. I removed the 0.8 version. Thank you @benjelloun

srchilukoori:
@benjelloun if you are ok with the current state of PR (after removing the version 0.8), please merge it.

PR #1036, opened 2026-06-01
Refactor visualizer script structure to modularize `visualize_js`

mqzhou-dev:
### **Motivation:**
Currently, `visualize.py` serves a dual role: it functions as a standalone executable CLI binary, but also exports `visualize_js` which is imported by `visualize_all.py`. 

To follow clean coding standards and maintain a modular architecture, this PR decouples the binary entrypoint from the library logic. By moving `visualize_js` and its helper functions into `visualize_utils.py`, we:
* Ensure that executable binaries (`visualize.py`, `visualize_all.py`) are lightweight CLI/application wrappers.
* House all core rendering and utility logic inside a dedicated library module (`visualize_utils.py`).
* Improve code reusability and prevent cross-binary import issues in strict compiler/packaging environments.

### **Testing:**
* Fully verified that the entire unit test suite passes successfully (`pytest . -m "not nonhermetic"`).
* Manually verified that both single-dataset rendering (`visualize`) and batch gallery rendering (`visualize_all`) generate identical HTML output pages as before the refactoring.


benjelloun (approved):
Thanks for the refactoring!

PR #1048, opened 2026-08-26
Compare hex checksums case-insensitively in Download._check_hash

karthikchundi-commits:
### What

`Download._check_hash` compares a downloaded file's SHA-256/MD5 hex digest against the `sha256`/`md5` value in the Croissant JSON-LD using a plain `==`. `hashlib`'s `hexdigest()` is always lowercase, but hex-encoded checksums are conventionally treated as case-insensitive by essentially every checksum tool (many emit uppercase, e.g. Windows' `Get-FileHash`). A dataset published with an uppercase hex checksum currently makes a correct, fully-intact download fail with a spurious `"Hash of downloaded file ... is not identical with the reference"` error.

### Fix

Lowercase only the hex comparison. The base64 fallback comparison right below it is intentionally left untouched - base64 is genuinely case-sensitive (upper/lowercase letters encode different values), so normalizing case there would risk silently accepting a corrupted file whose base64 hash happens to differ from the expected one only in letter case.

### Testing

Added `test_uppercase_hex_hash_does_match`, using the well-known SHA-256 of empty content (the same value the existing `test_sha256_hashes_do_match` already uses in lowercase), uppercased, across all three `CroissantVersion`s.

Ran the full `mlcroissant/_src/operation_graph/` test suite locally: 95 passed, 1 skipped, 3 pre-existing failures - all Windows path-separator (`\` vs `/`) comparison issues in unrelated functions (`test_get_download_filepath`, `test_get_fullpaths`, one `test_apply_transforms_fn` regex case), confirmed unrelated to this change by inspection; none touch `_check_hash` or the hash-comparison code path.


karthikchundi-commits:
recheck

